PR #2, opened 2015-11-14
Parse out Manufacturer Specific Data for API < 21

adolgov:
Manually parse out Manufacturer Specific Data from raw Advertising byte array when API < 21


smstuebe:
`i < rawScanRecord.Length` and `rawScanRecord[i + 1]` can cause IndexOutOfBoundsException, or?

adolgov:
Per BTLE spec, the advertisement is supposed to be LTV, so the
parsing shouldn't fail.

I suppose that devices could misformat their advertisements, especially
since BTLE is a little bit of a Wild West right now, so extra checks might
not be the worst thing in the world.

I guess the correct behavior in the case of a misbehaving peripheral is to
ignore the advertisement payload. I don't know if there is a current
mechanism for bubbling up errors associated with scan events and results.

xabre:
I will merge it like this for now. Won't release a nuget package until we check it :)

PR #3, opened 2015-11-19
iOS handling Device/Peripheral name change

adolgov:
Apparently, iOS hides the standard Generic Attribute service and its characteristics under the hood, and handles things like device name updates via events within the CBPeripheral.

Also changed Core project profile to 78 (from 111) since 78 is the current Xamarin "stable" (as far as I know) and I didn't see any 111-specific features being used. That one is mostly a personal preference, though.


xabre:
Great job :)

PR #58, opened 2016-06-30
Use the Client Characteristic Configuration Descriptor instead of the first available Descriptor

joechrysler:
The [BLE GATT Descriptors spec](https://developer.bluetooth.org/gatt/descriptors/Pages/DescriptorsHomePage.aspx) defines a specific Descriptor that clients should use to write [Client Characteristic Configurations](https://developer.bluetooth.org/gatt/descriptors/Pages/DescriptorViewer.aspx?u=org.bluetooth.descriptor.gatt.client_characteristic_configuration.xml) such as enabling notifications or indications. Although it is sometimes (often) the first descriptor that appears in `_nativeCharacteristic.Descriptors`, order isn't guaranteed. For some devices (including the one I'm working with now), another descriptor comes first and grabbing the first one out of the list will result in never actually enabling notifications on the characteristic.

### DIFF
--- a/Source/Plugin.BLE.Android/Characteristic.cs
+++ b/Source/Plugin.BLE.Android/Characteristic.cs
@@ -123,12 +123,10 @@ namespace Plugin.BLE.Android
 
             var successful = _gatt.SetCharacteristicNotification(_nativeCharacteristic, true);
 
-            // [TO20131211@1634] It seems that setting the notification above isn't enough. You have to set the NOTIFY
-            // descriptor as well, otherwise the receiver will never get the updates. I just grabbed the first (and only)
-            // descriptor that is associated with the characteristic, which is the NOTIFY descriptor. This seems like a really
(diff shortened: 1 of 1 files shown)


Source/Plugin.BLE.Android/Characteristic.cs
```diff
@@ -123,20 +123,19 @@ private bool InternalWrite(byte[] data)
...
+            // for details.
 
             await Task.Delay(100);
             //ToDo is this still needed?
             // HACK: did i mention this was a hack?????????? [CD] 50ms was too short, 100ms seems to work
 
             if (_nativeCharacteristic.Descriptors.Count > 0)
             {
-                var descriptor = _nativeCharacteristic.Descriptors[0];
+                var clientCharacteristicConfigurationDescriptorId = Java.Util.UUID.FromString("00002902-0000-1000-8000-00805f9b34fb");
```
joechrysler: I'm still a bit new to the project. There's probably a better place for this kind of BLE constant.

PR #59, opened 2016-06-30
Fix Aync Typo

onekam:
There were a few methods where "Async" was spelled "Aync." It isn't a terribly huge deal, but it became a bit of a hassle to remember which methods were spelled "Async" vs. "Aync" when my team and I were using the library.


xabre:
Currently merged it in our 1.0 branch.

PR #75, opened 2016-08-16
Added the ability to turn on/off bluetooth (Android only)

junian:


smstuebe:
Sorry, that's a bit too quick ^^ 

https://developer.android.com/reference/android/bluetooth/BluetoothAdapter.html#enable()

> Bluetooth should never be enabled without direct user consent. If you want to turn on Bluetooth in order to create a wireless connection, you should use the ACTION_REQUEST_ENABLE Intent, which will raise a dialog that requests user permission to turn on Bluetooth. The enable() method is provided only for applications that include a user interface for changing system settings, such as a "power manager" app.

Maybe we need 2 different methods. And we use spaces instead of tabs

adolgov:
On Android, I use the following in my BaseActivity (if your intention is to bug the person to turn on BT even if they click it off mid-activity). This is compatible with both Android 6.x+ and lesser versions. I vaguely recall that I had to change the implementation when moving to 6.

https://gist.github.com/adolgov/96121f4dbdbc475dfad257f7ed5f1254

I'm so sure that there's probably a way to include a gist inline, but alas, I'm not that great at GitHub-ing.

adolgov:
From what I remember, iOS does not allow changing BT state from within the app. I think that you're supposed to notify the user to turn it on, and then dump the user into the native iOS settings to finish the process.

flyingxu:
@adolgov I agree and I created issue #129

jopa7002:
Hello there, I'm trying to use your BLE plugin alongside a project that I'm working on and I would love to use your code. I'm having a hard time understanding how the plugin is suppose to work and what I need to do in order to establish and Adapter and BLE Scan and Find Advertisement Packets. I have gone through the documentation on the README and I've traced the code in the Sample App from Monkey.Robotics. However, the sample app is 3 years old and incompatible with Visual Studio 2017 so it's hard to understand how your library is used. I love that you guys made this open source and I would love to use your plugin within my project please feel free to e-mail and/or call or text me any time on my cell pjonson2@gmail.com (personal) payton.jonson@nielsen.com (work) 770-296-4712.

PR #109, opened 2016-11-16
TxPower is now parsed from advertising data for iOS

gtg529s:
TxPower is now parsed from advertising data for iOS.


xabre:
Great :)

PR #110, opened 2016-11-16
Master

gtg529s:
On further inspection of Android parsing of TxPower; value should be parsed as a single signed byte.  This commit makes the iOS and Android parsing return the identical value for TxPower.

### DIFF
--- a/Source/Plugin.BLE.iOS/Adapter.cs
+++ b/Source/Plugin.BLE.iOS/Adapter.cs
@@ -299,7 +299,11 @@ namespace Plugin.BLE.iOS
                 }
 				else if (key == CBAdvertisement.DataTxPowerLevelKey)
 				{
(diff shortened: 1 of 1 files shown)


gtg529s:
This fix allows TxPower returned from advertising data to be identical for iOS and Android.  TxPower can now be parsed for both platforms from the IDevice like such:
```
AdvertisementRecord ar = device.AdvertisementRecords.Where(x => x.Type == AdvertisementRecordType.TxPowerLevel).FirstOrDefault();
if (ar != null)
{
       //signed byte because TxPower can be negative (-100 to 20 to be exact)
	sbyte signedByte = unchecked((sbyte)ar.Data[0]);
	txPower = (int)signedByte;
}
```

smstuebe:
better:

     device.AdvertisementRecords.FirstOrDefault(x => x.Type == AdvertisementRecordType.TxPowerLevel);

Maybe add a check for length of `ar.Data` because `ar.Data[0]`

PR #148, opened 2017-02-01
Added a boolean (forceBleTransport) to force the use of 'transport' parameter to BLE  in connectGatt() method in Android.

m-anthoine:
With forceBleTransport = true, we will use 'nativeDevice.ConnectGatt(Application.Context, autoconnect, _gattCallback, BluetoothTransports.Le);' to force BLE. Very important for Dual Mode Device.

Sometimes with dual mode devices, we have to use the LE transport in the connectGatt method.
Otherwise the device will not respond to a create connection request (legacy).
The device will only respond to a create LE connection.
By default the transport mode is AUTO and in most case Android will decide to use BR/EDR connection (legacy)
The following is a proposition of how handle it by default.
If the device advertise for the 'Simultaneous LE and BR/EDR to Same Device Capable' capability (DualMode), the default transport mode should be LE in most of context.
For now, just add a boolean to force BLE transport in connectToDevice methods.


Source/Plugin.BLE.Android/Adapter.cs
```diff
@@ -154,12 +156,62 @@ protected override void StopScanNative()
...
+            //sometimes with dual mode devices, we have to use the LE transport in the connectGatt method.
+            //otherwise the device will not respond to a create connection request (legacy)
+            //the device will only respond to a create LE connection
+            //by default the transport mode is AUTO and in most case Android will decide to use BR/EDR connection (legacy)
+            //the following is a proposition of how handle it by default.
+            //if the device advertise for the 'Simultaneous LE and BR/EDR to Same Device Capable' capability (DualMode)
+            //the default transport mode should be LE in most of context
+            //For now, just add a boolean to force BLE transport in connectToDevice methods
+
+            //bool isLeBrErdCapableHost = nativeDevice.Type == BluetoothDeviceType.Dual;
```
smstuebe: can you please remove commented code?

Source/Plugin.BLE.Android/Adapter.cs
```diff
@@ -154,12 +156,62 @@ protected override void StopScanNative()
...
+
+            #endregion
+
+            if (forceBleTransport /* || isLeBrErdCapableHost*/)
```
smstuebe: commented code?

Source/Plugin.BLE.Abstractions/Contracts/IAdapter.cs
```diff
@@ -85,10 +85,11 @@ public interface IAdapter
         /// <param name="device">Device to connect to.</param>
         /// <param name="autoconnect">Android only: Automatically try to reconnect to the device, after the connection got lost. The default is false.</param>
         /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is None.</param>
+        /// <param name="forceBleTransport">Android only: For Dual Mode device, force transport mode to LE. The default is false.</param>
         /// <returns>A task that represents the asynchronous read operation. The Task will finish after the device has been connected successfuly.</returns>
         /// <exception cref="DeviceConnectionException">Thrown if the device connection fails.</exception>
         /// <exception cref="ArgumentNullException">Thrown if the <paramref name="device"/> is null.</exception>
-        Task ConnectToDeviceAsync(IDevice device, bool autoconnect = false, CancellationToken cancellationToken = default(CancellationToken));
+        Task ConnectToDeviceAsync(IDevice device, bool autoconnect = false, CancellationToken cancellationToken = default(CancellationToken), bool forceBleTransport = false);
```
smstuebe: I somehow think the CancellationToken should be the last parameter.
Should we introduce something like `ConnectParameter` with autoconnect and forceBleTransport  ? I don't like to have 20 platfom specific parameters in the interface.

Source/Plugin.BLE.Android/Adapter.cs
m-anthoine: Yes, I can remove it ;)
The purpose of this comment was to open a discussion on if we could apply a sort of default mecanism to auto enable the use of ble transport parameter.
I'm personaly faced to a device in dual mode wich not responds at all to a legacy connection (BR/ERD), but I don't know if it's the case for all dual mode devices :-/
What do you think of that ?

Source/Plugin.BLE.Android/Adapter.cs
m-anthoine: See my previous comment

Source/Plugin.BLE.Abstractions/Contracts/IAdapter.cs
m-anthoine: It seems to me that it's a good idea.
Would you like that i make a propostion for it ?

smstuebe: Yes, would be nice, if you can do that. It would reduce the refactoring effort on our side :)

Source/Plugin.BLE.Android/Adapter.cs
xabre: So the purpose of isLeBrErdCapableHost is to force a Dual device to always take into account this additional parameter. 
Without this flag an if the parameter is left to it's default value (false) a Dual device might be unpredictable and connect either in LE mode or BR/EDR?

m-anthoine: Just I was wondering, if ALL Dual devices work as mine, if it was better to automatically use the BLE transport parameter. So that's why I'm not sure of this feature and in doubt I prefered to implement a boolean to explicitly set the behavior. If someone knows better than me on the subject, I think it's an relevant question.

smstuebe:
@m-anthoine: Is it possible for you to change the target branch to 1.2.0 without creating a new PR?

m-anthoine:
@smstuebe I changed the base branch to 1.2.0.
Is it right for you ?
I'm not very aware about this kind of tasks ^^'

smstuebe:
@m-anthoine: Thanks.
@xabre: `Conflicting files Resource.Designer.cs` We checked this in just for bitrise, didn't we? Do you know if bitrise can handle this code generation, now?

m-anthoine:
@smstuebe I have juste resolved this conflict sorry but maybe @xabre can't see it now ^^'

PR #157, opened 2017-02-23
Add support for NSString descriptor values

simontreny:
IDescriptor.ReadAsync() fails (i.e. it returns null) on iOS if the descriptor's value is a string. This commit fixes this by converting the NSString value to an UTF8-encoded byte-array.

### DIFF
--- a/Source/Plugin.BLE.iOS/Descriptor.cs
+++ b/Source/Plugin.BLE.iOS/Descriptor.cs
@@ -28,6 +28,11 @@ namespace Plugin.BLE.iOS
                     return BitConverter.GetBytes(((NSNumber)_nativeDescriptor.Value).UInt64Value);
                 }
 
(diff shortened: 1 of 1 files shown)


Source/Plugin.BLE.iOS/Descriptor.cs
```diff
@@ -28,6 +28,11 @@ public override byte[] Value
                     return BitConverter.GetBytes(((NSNumber)_nativeDescriptor.Value).UInt64Value);
                 }
 
+                if (_nativeDescriptor.Value is NSString)
+                {
+                    return System.Text.Encoding.UTF8.GetBytes(((NSString)_nativeDescriptor.Value).ToString());
```
xabre: is the NSString::ToString() method not enough for the conversion?

xabre:
Nice :+1:

PR #167, opened 2017-03-12
Typo Fix

nithin-raj:
Typo fix from aync to async at few places


smstuebe:
thanks for multiple "s" 👍

Issue #128, opened 2016-12-19
Writing/reading >20 bytes


gaminn:
Hi,
is reading/writing >20 bytes from/to characteristic supported with this plugin? If so, please share how to do it?

xabre:
@gaminn Not yet :) 

We have an implementation idea (but we need more info) but nothing in the plugin yet:
- On android: use the RequestMtu call/callback 
- On iOS: not sure? open question?

Cheesebaron:
It seems to work fine for me? I just provide a byte array with the data and it eats it no problem. On Android I have had it eat around 450 bytes no problem.

sarkreth:
@Cheesebaron Are you sure the byte array makes it to the peripheral as expected? Also, what device are you using?

Cheesebaron:
Pretty sure it does. Have tried this on both iPhone and Android.

I can test again when I am back at work.

martinwiboe:
Can confirm that it only works for me if I call `RequestMtu` with a sufficiently large value prior to writing. I'm using a OnePlus 3 with Nougat.

xabre:
@martinwiboe Did you also try it on iOS?

martinwiboe:
@xabre No, not using your library, unfortunately. We do have a native iOS app that uses BLE to talk to the same product, and that works correctly without any adjustment to MTU etc.

martinwiboe:
@Cheesebaron have you had time to test this? :)

Dragon160:
We did an airtrace and discovered that iOS automatically calls an Exchange MTU request after establishing the connection. 
Android does not show this behaviour. I think the "requestMTU(xxx)" call needs to be done by the application itself.

@xabre Any plans to adopt this? :-)

PR #195, opened 2017-04-24
Added requestMTU feature (issue #128)

Dragon160:
Hi,

I added the possibility to request an MTU update for a peripheral. 
The requestMTU(int size) method can be called on an iDevice.  After calling it the function will return the new negotiated MTU (which can be differ from your requested one).

Feature Reason:
For iOS the bluetooth "Exchange MTU Request" will be called automatically after connection process. The MTU will be negotiated to about 158bytes for newer devices. Unfortunately on Android this is not done by itself..

There are a few restrictions:
Android < API 21: No such api available. An default value will be returned (-1)
iOS: Not possible to change MTU. The current negotiated MTU will be returned (CBPeripheral.GetMaximumWriteValueLength).

### DIFF
--- a/Source/BLE.Client/BLE.Client/ViewModels/DeviceListViewModel.cs
+++ b/Source/BLE.Client/BLE.Client/ViewModels/DeviceListViewModel.cs
@@ -399,6 +399,10 @@ namespace BLE.Client.ViewModels
                 {
                     _userDialogs.ShowLoading($"Connecting to {item.Name} ...");
                     await Adapter.ConnectToDeviceAsync(item.Device);
+
+                    var resultMTU = await item.Device.RequestMtuAsync(60);
+                    System.Diagnostics.Debug.WriteLine($"Requested MTU. Result is {resultMTU}");
+
                     item.Update();
                     _userDialogs.ShowSuccess($"Connected {item.Device.Name}");
 
--- a/Source/BLE.Client/BLE.Client/ViewModels/DeviceListViewModel.cs
+++ b/Source/BLE.Client/BLE.Client/ViewModels/DeviceListViewModel.cs
@@ -400,6 +400,7 @@ namespace BLE.Client.ViewModels
                     _userDialogs.ShowLoading($"Connecting to {item.Name} ...");
                     await Adapter.ConnectToDeviceAsync(item.Device);
 
+                    // TODO make this configurable
                     var resultMTU = await item.Device.RequestMtuAsync(60);
                     System.Diagnostics.Debug.WriteLine($"Requested MTU. Result is {resultMTU}");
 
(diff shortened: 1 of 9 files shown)


smstuebe:
Please spaces everywhere not Tabs

Source/Plugin.BLE.Android/Device.cs
```diff
@@ -221,5 +224,25 @@ public static List<AdvertisementRecord> ParseScanRecord(byte[] scanRecord)
...
+            {
+                Trace.Message($"Request MTU not supported in this Android API level");
+                return -1;
+            }
+
+            return await TaskBuilder.FromEvent<int, EventHandler<MtuRequestCallbackEventArgs>>(
+                execute: () => { _gatt.RequestMtu(requestValue); },
+                getCompleteHandler: (complete, reject) => ((sender, args) =>
+               {
+                   complete(args.Mtu);
```
xabre: Here you should also check for args.Error and if not null call reject(error)

Source/BLE.Client/BLE.Client/ViewModels/DeviceListViewModel.cs
```diff
@@ -399,6 +399,10 @@ private bool CanConnectToPrevious()
                 {
                     _userDialogs.ShowLoading($"Connecting to {item.Name} ...");
                     await Adapter.ConnectToDeviceAsync(item.Device);
+
+                    var resultMTU = await item.Device.RequestMtuAsync(60);
+                    System.Diagnostics.Debug.WriteLine($"Requested MTU. Result is {resultMTU}");
+
```
xabre: This will modify the sample app to always attempt to negotiate MTU... maybe we can make this configurable. We will add a ToDo

Source/Plugin.BLE.Android/GattCallback.cs
```diff
@@ -174,6 +176,14 @@ public override void OnReliableWriteCompleted(BluetoothGatt gatt, GattStatus sta
             Trace.Message("OnReliableWriteCompleted: {0}", status);
         }
 
+        public override void OnMtuChanged(BluetoothGatt gatt, int mtu, GattStatus status)
+        {
```
xabre: Because this is based on the current master and not on 1.3.0 we will need to also check device ID here. I.e the gatt callback instance is shared among all connected devices and so we need a way to identify the correct device. 
Maybe you can base your PR on the 1.3.0 and that we we will avoid this issue. (FYI 1.3.0 does some gatt callback refactoring in the sense that we have a callback per device instead of a global one). What do you think?

Source/Plugin.BLE.Android/Device.cs
xabre: Use the reject from here `getCompleteHandler: (complete, reject)`

xabre:
Have to merge master to 1.3.0 before we can rebase this PR on 1.3.0

xabre:
Available starting with 1.3.0-beta1

Dragon160:
Great 🤘

PR #200, opened 2017-04-28
Added possibility to change ConnectionPriority/ ConnectionInterval for Android

Dragon160:
Hi,

The Android BLE stack is able to define the connection priority (high, low, balanced) and therefore change the connection interval for a connection.
The connection interval influences the ble performance and battery consumption massively.

**Android Gatt Function:**
https://developer.android.com/reference/android/bluetooth/BluetoothGatt.html#requestConnectionPriority(int)

**Android Settings for Connection Priority:**
https://developer.android.com/reference/android/bluetooth/BluetoothGatt.html#CONNECTION_PRIORITY_BALANCED

Its not available on iOS so this function will only work for Android (and API Level >= 21).
On iOS the call will return TRUE. Why true? I think returning FALSE could lead to unexpected behaviour if somebody reacts on the return value in the consuming app.
Please comment on this..

### DIFF
--- a/Source/Plugin.BLE.iOS/Device.cs
+++ b/Source/Plugin.BLE.iOS/Device.cs
@@ -135,5 +135,11 @@ namespace Plugin.BLE.iOS
             Trace.Message($"Request MTU is not supported on iOS.");
             return await Task.FromResult((int)_nativeDevice.GetMaximumWriteValueLength(CBCharacteristicWriteType.WithoutResponse));
         }
+
+        protected override async Task<bool> UpdateConnectionIntervalNativeAsync(ConnectionInterval interval)
+        {
+            Trace.Message("Cannot update connection inteval on iOS. Returning TRUE ..");
+            return await Task.FromResult(true);
+        }
     }
 }
--- a/Source/Plugin.BLE.iOS/Device.cs
+++ b/Source/Plugin.BLE.iOS/Device.cs
@@ -136,10 +136,10 @@ namespace Plugin.BLE.iOS
             return await Task.FromResult((int)_nativeDevice.GetMaximumWriteValueLength(CBCharacteristicWriteType.WithoutResponse));
         }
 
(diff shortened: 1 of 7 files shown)


Source/Plugin.BLE.iOS/Device.cs
```diff
@@ -135,5 +135,11 @@ public void Update(CBPeripheral nativeDevice)
             Trace.Message($"Request MTU is not supported on iOS.");
             return await Task.FromResult((int)_nativeDevice.GetMaximumWriteValueLength(CBCharacteristicWriteType.WithoutResponse));
         }
+
+        protected override async Task<bool> UpdateConnectionIntervalNativeAsync(ConnectionInterval interval)
+        {
+            Trace.Message("Cannot update connection inteval on iOS. Returning TRUE ..");
+            return await Task.FromResult(true);
```
smstuebe: return true; when async
or
return Task.FromResult(true); when not async. I'd prefer this one, because it doesn't generate a state machine.

Question: Why not false? False would be better, or? If you compare it to the Android implementation.

Dragon160: According to async: I will change it to "return true".

About the return value:
Hm I was not sure about the expected behavior from the users perspective.... maybe "false" would be better since its not possible to apply anything; I agree!

xabre:
Cool :) Just one question, why did you design the api method to be async if none of the platform specific implementations are in fact async. I'm guessing this can drop the async part, or am I missing something?

Dragon160:
Actually I was not really sure if the function internally needs to call any other async functions (now or in the future) so I decided to make the API async for the caller.

xabre:
I would suggest that we follow the YAGNI principle. If Android or iOS change their API's they will have breaking cahnges and we will adapt to it. I'd keep it as simple as possible :)

PR #219, opened 2017-06-16
Fixed potential deadlock while discovering services

JSchaenzle:
The handler function can be called for two reasons: 1) service discovery succeeded or 2) service discovery failed. In the later case `args.Error` will not be null and `_nativeDevice.Services` may be null. In that case, we cannot return from the handler without setting the result on the TaskCompletionSource. Doing so will certainly cause a deadlock because this task will never complete. The fact that the event handler is removed by `_nativeDevice.DiscoveredService -= handler;` ensures that. 

If an error occurs, the code will not complete with an empty list of services. Then the caller can do what they choose with that.

### DIFF
--- a/Source/Plugin.BLE.iOS/Device.cs
+++ b/Source/Plugin.BLE.iOS/Device.cs
@@ -54,16 +54,16 @@ namespace Plugin.BLE.iOS
                 {
                     Trace.Message("Error while discovering services {0}", args.Error.LocalizedDescription);
                 }
-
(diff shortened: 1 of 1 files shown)


smstuebe:
Thanks for the PR!

I think we should directly throw an exception, if the native api returns an error. 

What do you think @xabre and @JSchaenzle ?

xabre:
@smstuebe can you please try to run the ios sample app anc check if service discovery is still working? i don't have an apple developer account at the moment :( 
danke :)

Issue #227, opened 2017-07-07
StopUpdateAsync does not write the characteristic's descriptor to disable notification


bed007:
The Android implementation of StopUpdateAsync is not writing the ClientCharacteristicConfigurationDescriptor to `BluetoothGattDescriptor.DisableNotificationValue` so the peripheral is never told that the notification should be disable.

## Steps to reproduce

1. call StartUpdateAsync() on a ICharacteristic

2. call StopUpdateAsync() on the same ICharacteristic

3. check the peripheral device, it is still performing the notify updates

## Expected behavior
The StopUpdateAsync should, just like the StartUpdateAsync does, write to the ClientCharacteristicConfiguration descriptor to disable the notification.

## Suggested fix
Quick fix using #if __ANDROID__ after my call to StopUpdateAsync:

```
var chr = await GetCharacteristic(device, svcId, chrId);

chr.ValueUpdated -= Characteristic_ValueUpdated;
await chr.StopUpdatesAsync();

#if __ANDROID__
// This is a fix for a bug in Plugin.BLE that does not write Notify Disable in the descriptor
var descriptor = await chr.GetDescriptorAsync(ClientCharacteristicConfigurationDescriptorId);

await descriptor.WriteAsync(Android.Bluetooth.BluetoothGattDescriptor.DisableNotificationValue.ToArray());
#endif
```

## Configuration

**Version of the Plugin:** 1.2.3

**Platform:** Android 7.1 (API Lvl 25)

**Device:** Google Nexus 6P

smsissuechecker:
Hi @bed007,

I'm the friendly issue checker.
It seems like (25.00 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

If you think it's fine to make an exception, just ignore this message.
**But if you think it was a mistake to delete the template, please close the issue and create a new one.**

Thanks!

bed007:
hello smsissuechecker, I believe I've followed it very much... I don't agree with your analysis ;-)

smstuebe:
its just a 25% confidence :P If you want, you can provide a more intelligent approach: http://smstuebe.de/2017/01/12/github-issue-checker-azure-function/

xabre:
Merged PRs :) Thank you.

Issue #198, opened 2017-04-27
Cached services and characteristics not cleared on signal loss


xabre:
On Android, cached services are not cleared on signal loss.
Services are cleared only when the user triggers the device disconenct and not when disconnection is caused by something else.

## Steps to reproduce

1. Connect to device.

2. Trigger a signal loss.

3. Get back in range.

4. Reconnect using same IDevice instance. 

5. Service/ Characteristic operations will fail because they work on stale cached instances. For example Characteristic write / read.

## Expected behavior
Clear cached services/characteristic on any type of connection loss.

## Actual behavior
Cached Services/Characteristics etc not cleared

### Crashlog

Not applicable

## Configuration

**Version of the Plugin:** 
1.3.0-beta1

**Platform:** 
Android All

**Device:** 
All android devices.

smsissuechecker:
Hi @xabre,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

xabre:
Fixed with 1.3.0-beta2

PR #234, opened 2017-07-19
[Android] 1.3.0 - GATT operations issue on reconnecting on a previously connected IDevice

bed007:
This is a fix for Android when reconnecting to a previously connected IDevice, any GATT operations would return FALSE because the services & characteristics cache is no longer valid.

I am simply calling DeviceBase.ClearServices() (as it is in iOS) when device is disconnected (lost or intentionnaly).

I've also merge my PR for Issue #227 into the 1.3.0 branch.


xabre:
@bed007 Thanks for the PR. 
The clear services method was already fixed with 1.3.0-beta2. #198

Issue #182, opened 2017-03-31
Autoconnect on Android 


xabre:
Investigate and document the behaviour of Android the auto-connect flag.

smsissuechecker:
Hi @xabre,

I'm the friendly issue checker.
It seems like (100.00 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

If you think it's fine to make an exception, just ignore this message.
**But if you think it was a mistake to delete the template, please close the issue and create a new one.**

Thanks!

AnoopChatra:
Hi ,

    I am not at all able to re-connect  to the BLE device using xamarin-bluetooth-le plugin, though "auto-connect" flag is set to true. I have tested my BLE devices with the "native android mobile apps" and "mono android mobile apps" the  auto-connect / auto-reconnect works perfectly fine in either of the cases.

In xamarin-bluetooth-le plugin I am using the following method to connect to the BLE device and I have successfully connected:
CrossBluetoothLE.Current.Adapter.ConnectToDeviceAsync(deviceToBeConnected,new ConnectParameters(true,false)) ;
Here the first param of ConnectParameters is "AutoConnect" flag , which needs to be set to true for auto re-connections.

Inside ConnectToDeviceAsync() method,  ConnectGatt() will be called: 
((BluetoothDevice)device.NativeDevice).ConnectGatt(Application.Context, connectParameters.AutoConnect, _gattCallback) 
Here _gattCallback is the implementation of BluetoothGattCallback. It contains overridden method OnConnectionStateChange() which gets the BLE connection change status.

Inside  OnConnectionStateChange() on ProfileState.Disconnected " gatt.Close();" is getting called.When close() is called I believe the "gatt" object will be deallocated and it wont be available anymore .So there is no chance of re-connection.

Is it really required to call  " gatt.Close();" on  ProfileState.Disconnected change ? Is there any reason behind it? 
I guess auto-connection will work if  we do not close the Gatt object. Could you please help me out to solve this issue ?? 

Regards,
Anoop Chatra

AnoopChatra:
Hey @xabre ,

      Any updates about auto connect??

AnoopChatra:
Hi @xabre

   I have pulled the latest code which  contains the fix for "auto reconnection" in android. I have built the dlls locally and I am using it in my application.
[...]

PR #242, opened 2017-08-11
#182 Android auto connect: Modify Connection Parameters To Allow A Developer To Bypass Autoconnect

jopa7002:
The purpose of this pull request is to implement the following features. 

Expand the ConnectParameters.cs and the Device.cs file to allow a developer to pass a bluetooth pairing pin to a bluetooth peripheral from an Android device.

Expand the ConnectParamaters.cs and the Device.cs file to allow a developer to bypass the implemented Android auto connect and prompt the Android OS to ask the user for the pairing pin. Once the user correctly enters the pairing pin the bluetooth connection will be made by the Android OS.


janusw:
This PR is old and very outdated by now (with a multitude of conflicts). If there is still interest in this feature, a new PR should be opened on top of on the current master branch. I'll close this one.

PR #247, opened 2017-08-28
[UWP]Use Adapter.ScanMode to set UWP's  ScanningMode

flyingxu:
During my test, ~~if `ScanningMode = BluetoothLEScanningMode.Active` I got no device scanned.
But if `ScanningMode = BluetoothLEScanningMode.Passive` (default value), I got device list same as the result of Android/iOS.~~
EDIT: I tested again the other day, the Thinkpad E450 can scan my devices at `Active` mode. 

Another example is from https://github.com/Microsoft/Windows-universal-samples/blob/master/Samples/BluetoothAdvertisement/cs/Scenario1_Watcher.xaml.cs and it uses the default (`Passive`) value.

### DIFF
--- a/Source/BLE.Client.UWP/Package.appxmanifest
+++ b/Source/BLE.Client.UWP/Package.appxmanifest
@@ -45,5 +45,6 @@
 
   <Capabilities>
     <Capability Name="internetClient" />
+    <DeviceCapability Name="bluetooth" />
   </Capabilities>
 </Package>
--- a/README.md
+++ b/README.md
@@ -52,6 +52,14 @@ Add this line to your manifest if you want to declare that your app is available
 <uses-feature android:name="android.hardware.bluetooth_le" android:required="true"/>
 ````
 
+**UWP**
+
+Add this line to the Package Manifest (.appxmanifest):
+
+```xml
+<DeviceCapability Name="bluetooth" />
+```
+
 ## Sample app
 
 We provide a sample Xamarin.Forms app, that is a basic bluetooth LE scanner. With this app, it's possible to 
--- a/Source/BLE.Client.UWP/BLE.Client.UWP.csproj
+++ b/Source/BLE.Client.UWP/BLE.Client.UWP.csproj
@@ -133,7 +133,7 @@
       <Version>6.5.1</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
-      <Version>5.2.3</Version>
+      <Version>5.4.0</Version>
     </PackageReference>
     <PackageReference Include="MvvmCross">
       <Version>5.1.0</Version>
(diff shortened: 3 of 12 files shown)


flyingxu:
Updated to use Adapter.ScanMode.

smstuebe:
A bit quick :) 
The `ScanMode` defaults to `LowPower` wich is now mapped to `Active` and  that means your original PR intention is lost.

I've these questions:
- Makes my mapping sense? You are the UWP expert :P 
-  You say, you do not find any devices. Why is that?

flyingxu:
As long as there is a way to set the `ScanMode` in my client code (ViewModel), it will be ok for me. So the default value of `LowPower` does not matter to me because I can change it in my own code after the PR.

I'm not a UWP expert but you asked great questions. After a rough search on the internet and I found no way to set Android scan mode to passive, so maybe all the Android scan modes should be mapped to `Active` ?  Back to your current mapping, do you have any supporting code, like Android source code?

Actually, Android's implementation (no API to set to be Passive) makes sense, because `Active` covers `Passive` according to Bluetooth spec **in theory**. I'm not sure why my Android phone can scan my devices in `Active` mode (if my previous assumption is right) but my UWP PC can only scan in `Passive` mode. I would guess it's a UWP PC driver implementation issue. I will test on different PC again.

flyingxu:
My other PC (YOGA 900, Win10 15063) can scan devices at **both** `Passive` and `Active` mode

flyingxu:
Just keep as a reference: https://github.com/aritchie/bluetoothle/blob/563fca9e5e40320577545e5dba1f010d86c7c264/Plugin.BluetoothLE.Uwp/AdapterContext.cs#L53 uses Active always.

smstuebe:
Hmm we actually implemented it in Android. This was requested by Cheesebaron, because he wanted high energy scanning.

https://github.com/xabre/xamarin-bluetooth-le/blob/b30950113e40defcf1d612fdc972708be2137a16/Source/Plugin.BLE.Android/Adapter.cs#L108-L109

> My other PC (YOGA 900, Win10 15063) can scan devices at both Passive and Active mode

Hmm then I think the mapping for UWP is ok!?

Source/BLE.Client.UWP/Package.appxmanifest
```diff
@@ -45,5 +45,6 @@
 
   <Capabilities>
     <Capability Name="internetClient" />
+    <DeviceCapability Name="bluetooth" />
```
smstuebe: We should document this in the readme.

flyingxu: Done

flyingxu:
I updated my issue to make a correction of my UWP test result, it can scan devices at `Active` mode, so my issue is gone.

But this PR is still valid and I edited the title of this PR to reflect it.

xabre:
@smstuebe Are you planning to merge these changes?

PR #248, opened 2017-09-08
Typos / Null check / Library updates / Fixed connection status events

fozzzgate:
I submitted these pull requests to flyingxu's fork and he accepted them, but said that it would be better if I submitted them directly to the base repository


xabre:
@fozzzgate Great job :) Thank you :)

Issue #301, opened 2018-05-03
Update to MvvmCross 6


fonix232:
The latest major release of MvvmCross, version 6.0.0 (well, technically, 6.0.1) was just released. It changes some stuff, most importantly, the countless base libraries were merged into a single assembly.

Unfortunately this project still depends on the old libraries, making it near impossible to be used with 6.0.

The library should be updated to .Net Standard and MvvmCross 6.x

smsissuechecker:
Hi @fonix232,

I'm the friendly issue checker.
It seems like (100.00 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

If you think it's fine to make an exception, just ignore this message.
**But if you think it was a mistake to delete the template, please close the issue and create a new one.**

Thanks!

flyingxu:
@fonix232 Thanks, I'd like to see why you say it's near impossible to be used with MvvmCross 6.0

flyingxu:
closed by #307

Issue #285, opened 2018-02-23
.NETStandard v2.0 support


dinodzakmic:
By creating new Xamarin.Forms project based on .NETStandard 2.0, Plugin.BLE packages are restored using .NETFramework v4.6.1. 

Rebuilding the solution inside VisualStudio 2017 is successfull, but if I try to set up Jenkins or VS Mobile center for CI/CD/CT, then rebuild fails because of this "issue".
Inside Jenkins I am trying to build using MSBUILD.

Is there any workaround or do you soon plan to add .NETStandard 2.0 support?

smsissuechecker:
Hi @dinodzakmic,

I'm the friendly issue checker.
It seems like (100.00 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

If you think it's fine to make an exception, just ignore this message.
**But if you think it was a mistake to delete the template, please close the issue and create a new one.**

Thanks!

biapar:
Do it work with .NETStandard 2.0?

dinodzakmic:
@biapar  It's working when I use my machine for build and deployment, but on my Jenkins server it does not work.

biapar:
Which is the min .NET version supported?

smstuebe:
its using PCL profile 259 which is netstandard1.0. netstandard1.0 is a subset of netstandard2.0. So it will work.

flyingxu:
Plugin.BLE (2.0.0-pre1) is .NETStandard 1.0
```
NETStandard.Library (>= 1.6.1) 
```

PR #303, opened 2018-05-17
Mvvmcross6 xf2

flyingxu:
the MvvmCross package is updated to v6, for #301 
-Navigation change
-plugin load change
-MvxContentPage change
-Log change
-Acr.UserDialogs change

 The PLC projects are also updated to NetStandard 2 projects, partially addresses #285 . The files structure is not updated yet.


flyingxu:
will submit another PR

PR #305, opened 2018-05-17
Update the Mvvmcross package to v6 and use Xamarin.Froms v2

flyingxu:
the MvvmCross package is updated to v6, for #301
-Navigation change
-plugin load change
-MvxContentPage change
-Log change
-Acr.UserDialogs change

The PLC projects are also updated to NetStandard 2 projects, partially addresses #285 . The files structure is not updated yet.

This PR use Xamarin.Forms v2.5 purposely, because XF v3.0 has a bug currently.


smstuebe:
All this new stuff should be considered as 2.0.0.
We should create a branch for that, so we can introduce breaking changes. James did a new major release for the switch to netstandard, and I will do it, too for my fingerprint plugin.

flyingxu:
Sure, I have no objection to that.

xabre:
I'm assuming we can re-target the PR to 2.0.0

smstuebe:
https://blog.nuget.org/20140924/supporting-semver-2.0.0.html
now, that nuget supports semver, finally, maybe `-pre.1` would have been the better choice :D

xabre:
Good point. We can name the next one -pre1.1 :))

flyingxu:
close this for a new PR to the branch of xabre:2.0.0

PR #307, opened 2018-05-29
upgrade branch 2.0.0 to use mvvmcross6

### LINKED ISSUE #301: Update to MvvmCross 6

flyingxu:
Upgraded the project to MvvmCross 6.

Because the bootstrap file is not needed in MvvmCross 6, `MvvmCross.Plugins.BLE` is now an empty project, but it has a reference of `Plugin.BLE.Abstractions` now. It has no dependency to MvvmCross any longer. 
Nuget spec file should be updated but not in this PR. 
This project can be removed entirely and the user can use Plugin.BLE directly but not in this PR.

`MvvmCross.Plugins.BLE` iOS and Droid are using the new MvvmCross 6 plugin loader class. On iOS, this class should be linked in LinderPleaseLink file. 

`BLE.Client` is recreated to be a netstandard 2 project to use MvvmCross 6 page navigation.
Splat is upgraded to v4 to work with netstandard 2.
Acr.UserDialogs is also upgraded to use netstandard. 

This should fix #301


xabre:
Looking over the code, it looks fine. I tried to run the android sample however and had a bunch of issues building it. For one the XF dependecy on the android support libs does not match the dependency from Acr.UserDialogs. Anyway not straight forward, will update my environment tonight and try again.

flyingxu:
Please do so and have another try.

I can confirm that the android sample works fine here on both Windowns and Mac.

fonix232:
The sample compiles and runs fine for me too. However for some reason I can't archive it (might be just my setup, though).

IMO the sample project should not depend on anything extra but the Android base libraries, MvvmCross, MvvmCross.Plugin.BLE and maybe the support libraries, but that's it.

fonix232:
@xabre had any luck trying to build it again?

xabre:
Yes, it's all fine :) Thanks for being patient :)

flyingxu:
@fonix232 I created #316 and you can follow the instructions to build you local nuget file.

PR #314, opened 2018-06-19
[iOS] Initial support for state restoration with CentralManagerDelegate

genriquez:
Use `CentralManagerDelegate` instead of `CentralManager` events for handling BLE events on iOS. This will allow overrides of the default `BleImplementation`, so you can pass a `RestorationIdentifier` to `CentralManager` and keep all pending connections alive after the app was terminated due to memory pressure.

`BleCentralManagerDelegate` maps all the method overrides to the same events from the `CentralManager` to minimize code changes.

Example of `BleImplementation` override to pass a `RestorationIdentifier`:
```
protected override void InitializeNative()
{
    var cmOptions = new CBCentralInitOptions { RestoreIdentifier = BluetoothStateName };
    var cmDelegate = new BleCentralManagerDelegate();
    CentralManager = new CBCentralManager(cmDelegate, DispatchQueue.CurrentQueue, cmOptions);

    _centralManagerDelegate = cmDelegate;
    _centralManagerDelegate.UpdatedState += (s, e) => State = GetState();
}
```

Details on the iOS requirements for state restoration can be found [here](https://developer.apple.com/library/archive/documentation/NetworkingInternetWeb/Conceptual/CoreBluetooth_concepts/CoreBluetoothBackgroundProcessingForIOSApps/PerformingTasksWhileYourAppIsInTheBackground.html#//apple_ref/doc/uid/TP40013257-CH7-SW10).

### DIFF
--- a/Source/Plugin.BLE.iOS/Adapter.cs
+++ b/Source/Plugin.BLE.iOS/Adapter.cs
@@ -15,21 +15,24 @@ namespace Plugin.BLE.iOS
     {
         private readonly AutoResetEvent _stateChanged = new AutoResetEvent(false);
         private readonly CBCentralManager _centralManager;
+	    private readonly IBleCentralManagerDelegate _bleCentralManagerDelegate;
 
-        /// <summary>
-        /// Registry used to store device instances for pending operations : disconnect
-        /// Helps to detect connection lost events.
-        /// </summary>
-        private readonly IDictionary<string, IDevice> _deviceOperationRegistry = new ConcurrentDictionary<string, IDevice>();
+		/// <summary>
+		/// Registry used to store device instances for pending operations : disconnect
+		/// Helps to detect connection lost events.
+		/// </summary>
+		private readonly IDictionary<string, IDevice> _deviceOperationRegistry = new ConcurrentDictionary<string, IDevice>();
         private readonly IDictionary<string, IDevice> _deviceConnectionRegistry = new ConcurrentDictionary<string, IDevice>();
 
         public override IList<IDevice> ConnectedDevices => _deviceConnectionRegistry.Values.ToList();
 
 
-        public Adapter(CBCentralManager centralManager)
+        public Adapter(CBCentralManager centralManager, IBleCentralManagerDelegate bleCentralManagerDelegate)
         {
             _centralManager = centralManager;
-            _centralManager.DiscoveredPeripheral += (sender, e) =>
+	        _bleCentralManagerDelegate = bleCentralManagerDelegate;
+
+            _bleCentralManagerDelegate.DiscoveredPeripheral += (sender, e) =>
             {
                 Trace.Message("DiscoveredPeripheral: {0}, Id: {1}", e.Peripheral.Name, e.Peripheral.Identifier);
                 var name = e.Peripheral.Name;
@@ -45,13 +48,13 @@ namespace Plugin.BLE.iOS
                 HandleDiscoveredDevice(device);
             };
 
-            _centralManager.UpdatedState += (sender, e) =>
+            _bleCentralManagerDelegate.UpdatedState += (sender, e) =>
             {
                 Trace.Message("UpdatedState: {0}", _centralManager.State);
                 _stateChanged.Set();
             };
 
-            _centralManager.ConnectedPeripheral += (sender, e) =>
+            _bleCentralManagerDelegate.ConnectedPeripheral += (sender, e) =>
             {
                 Trace.Message("ConnectedPeripherial: {0}", e.Peripheral.Name);
 
@@ -74,7 +77,7 @@ namespace Plugin.BLE.iOS
                 HandleConnectedDevice(device);
             };
 
-            _centralManager.DisconnectedPeripheral += (sender, e) =>
+            _bleCentralManagerDelegate.DisconnectedPeripheral += (sender, e) =>
             {
                 if (e.Error != null)
                 {
@@ -107,7 +110,7 @@ namespace Plugin.BLE.iOS
                 HandleDisconnectedDevice(isNormalDisconnect, foundDevice);
             };
 
-            _centralManager.FailedToConnectPeripheral +=
+            _bleCentralManagerDelegate.FailedToConnectPeripheral +=
                 (sender, e) =>
                 {
                     var id = ParseDeviceGuid(e.Peripheral);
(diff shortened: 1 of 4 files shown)


xabre:
Hi @genriquez. Thank you for the contribution :) I will retarget the PR to our 2.0.0 branch as this is our current RC.

genriquez:
Thanks, I completely overlooked that branch. I see there are more usages of `CentralManager` events there. I'll update the PR with those changes as well.

smstuebe:
Do we need to add something to the readme or documentation (advanced topic) regarding the state restoration?

genriquez:
My objective with this was to at least get support for iOS to relaunch my app in background when a pending device connection is resolved. To do so, you need to use a custom `BleImplementation` class, with the `InitializeNative()` implementation highlighted in the PR description. This would mean not being able to use the `CrossBluetoothLE.Current` property, so if you think this should be part of the docs, it should be listed as a workaround for now.

The ideal scenario would be to implement support for this within the plugin, including support for the `WillRestoreState` event, which should be used to hydrate the plugin instance with the restored device instances. I'm currently not using that event for anything in my app, so I'm not 100% sure what kind of work needs to be done there. More info on the `WillRestoreState` event [here](https://developer.apple.com/library/archive/documentation/NetworkingInternetWeb/Conceptual/CoreBluetooth_concepts/CoreBluetoothBackgroundProcessingForIOSApps/PerformingTasksWhileYourAppIsInTheBackground.html#//apple_ref/doc/uid/TP40013257-CH7-SW14).

xabre:
@genriquez Thanks for the update, I agree that we need a platform specific method for configuration. 
I think it's a good idea to add the suggested static `BleImplementation.UseRestorationIdentifier(identifier)`. Maybe you can also update the Readme.md and add a new link to the **Extended topics** section with the explanations of how to use this ocnfiguration method.

genriquez:
Let me know if you need any more updates from my side.

samosim:
Hi Genriquez

I implemented the Bluetooth background  mode on ios in the same manner as you done, but I observed that background works fine one or at least two days after that the app is terminated. I use Xamarin form. Have you ever observed the same issue? It is worth mentioning that WillTerminated method is not called I only see that I do not get a local notification anymore.

I will appreciate for any your help.
Samo

Samo

genriquez:
Hi @samosim,

So the app stops working after 2 days? Just to be clear, we are talking about the app being terminated by iOS, not by the user, right?

We haven't seen any issues with this yet. You must make sure that the BLE connectivity logic does not depends on any UI implementation (like events from Xamarin Forms), since the app is relaunched in background, and the UI is not initialized until the user restores the application to the foreground.

This is just a way to keep BLE connection attempts alive after the app is killed due to memory pressure. If iOS decides to remove the app from memory, when a pending BLE connection is completed you should see an entry in the phone logs like this:

`✝🕆🕇 Resurrecting zombie application "your.package.name" 🕇🕆✝`

After this, the app executes the entry point in the AppDelegate, and from there you need to restart whatever process you were executing.

genriquez:
Ah, I see. I haven't used it with background scanning, only for direct connections to devices, so I don't have any specific experience for this use case. According to the [documentation](https://developer.apple.com/library/archive/documentation/NetworkingInternetWeb/Conceptual/CoreBluetooth_concepts/CoreBluetoothBackgroundProcessingForIOSApps/PerformingTasksWhileYourAppIsInTheBackground.html#//apple_ref/doc/uid/TP40013257-CH7-SW10), state preservation covers scanning.

Are you seeing the `Resurrecting zombie application` message? If you see it only once, and after that the app stops working, then I would suspect that when the app is restored in background, the `DidEnterBackground` event is not triggered. In my use case we always fire a Task from the `FinishedLaunching` event, which attempts to connect to devices. Take into account that the event of the discovered device might be lost, since the current implementation just allows the app to be relaunched, and ignores the restored state.

Other than that, I would try to check the logs for what is happening. You can install some heavy apps and use them intensively to force iOS to terminate your app, so you might not need to wait for days for the app to stop responding. Since you can reproduce the issue without using the BLE plugin, you might want to consider posting this on Stackoverflow as well.

samosim:
Thanks for your reply. If I am honest I do not know what does Resurrecting zombie application mean. Can you tell me a little bit more about this application?  It is interesting that my application works one or two days after that I do not get a local notification anymore. At the same time, the methods WillRestore and WillTerminate have not been triggered. 

Can you tell me what happens when the application is restored in the background? I assume that WillRestore method should be called. I do no understand why DidEnterBackground method should be called. I will appreciate your help.

Thanks in advance.

Samo

genriquez:
The `Resurrecting zombie application` is a message shown in the logs when your app is re-launched in background after a change in your pending BLE tasks:

image

Our use case was just waiting for a specific device to be available so we can connect to it. Pending connections are kept alive by iOS after the app is terminated, and relaunches it when the device becomes available. When relaunched, we attempt to connect again to the device.

If the app when re-launched does not execute `DidEnterBackground` for your scenario, then the scanning process is never restarted. When the app is re-launched after being terminated, it is similar to the user opening it from scratch, but without any UI events/initialization, and with a time limit for it to perform all the required work. You need to "manually" resume all relevant tasks in this scenario (in your case, the scanning).

samosim:
Genriquez, you are a really kind person. Thanks again for your dedicated time and help.  I added in FinishedLaunching method the local notification and when the app is re-launched the local notification is shown. It means that the FinishedLauching method is triggered as you said. At this point I have still one question regarding re-launching procedure. According to your explanation, All relevant tasks which are responsible for scanning have to be called manually. But I do not know how FinishedLaunching method will recognize that these tasks have to be performed only when re-launching is caused by iOS. In other words, how I can separate the code which has to be executed only in the case when the iOS re-launches the application and not user opens the application. In both cases, the FinishedLaunching method is called.

Thanks in advance and happy holidays.

Samo

genriquez:
I'm afraid this is as far as my knowledge goes. We implemented this after we came across memory pressure scenarios in our tests, and noticed that there is support from iOS to keep things running, and the plugin didn't implement the feature yet. According to the [documentation](https://developer.apple.com/library/archive/documentation/NetworkingInternetWeb/Conceptual/CoreBluetooth_concepts/CoreBluetoothBackgroundProcessingForIOSApps/PerformingTasksWhileYourAppIsInTheBackground.html#//apple_ref/doc/uid/TP40013257-CH7-SW10) it should work, although I read in some forums that you may need to scan with a specific service UUID filter for it to be persisted.

PR #316, opened 2018-06-23
Update the build / nuget pack

flyingxu:
Remove the .pp file
Add netstandard1.0 nuget
Update the README

### DIFF
--- a/.build/MvvmCross.Plugin.BLE.nuspec
+++ b/.build/MvvmCross.Plugin.BLE.nuspec
@@ -1,7 +1,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
   <metadata>
     <id>MvvmCross.Plugin.BLE</id>
-    <version>2.0.0-pre1</version>
+    <version>2.0.0-pre2</version>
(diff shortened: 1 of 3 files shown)


fonix232:
I think the whole nuget spec should be rewritten - right now the format and structure of the code is of the old style NuGet packaging with nuget.exe, etc. all bundled, but since nuget 3.0 this is not needed.

The other targets (win, uap, etc.) all should be updated to use the latest MvvmCross 6.0 reference, and I think the PCL target could also be dropped - MvvmCross 6 transitions to the .Net Standard style of stuff, and I believe this project should follow it too.

Right now I don't have access to a machine with Visual Studio so I can't really make the needed changes, but if this isn't merged by next weekend I'll take a look at it.

xabre:
Thanks :) I would also like ot migrate all build script to cake....sometime in the future ;)

flyingxu:
@fonix232 There is a mac build script: 

`xamarin-bluetooth-le\.build\build.sh`

fonix232:
@flyingxu I doubt I can run any of the build scripts on my Note8 :) Unfortunately I'm temporarily without computer.

PR #332, opened 2018-10-25
2.0.0-Pre2

flyingxu:
use MvvmCross 6.0.
Fix the build scripts,
Document the steps to build/pack the NuGet package locally.
Rebuild the MvxPluginTest solution to verify the pre2 plugins can be  loaded in iOS and Android


flyingxu:
I'd like to merge this so that we can move forward.

PR #358, opened 2019-05-24
Added an option to ShowPowerAlert for iOS.

### LINKED ISSUE #129: Add CBCentralInitOptions support
On iOS, currently `CBCentralManager` is initialized by 

`_centralManager = new CBCentralManager(DispatchQueue.CurrentQueue);`

https://github.com/xabre/xamarin-bluetooth-le/blob/e20941a7b9aeecec8d6d865652c7f76eae4bcc5b/Source/Plugin.BLE.iOS/BleImplementation.cs#L22

There is no way to pass in `CBCentralInitOptions` parameters. by `CBCentralInitOptions`, [system should display a warning dialog to the user if Bluetooth is powered off ](https://developer.apple.com/reference/corebluetooth/cbcentralmanageroptionshowpoweralertkey), then the user can be redirected to iOS setting, which is convenient for the user.

Thorinos:
Fixes #129 

User can decide to enable or not the option `ShowPowerAlert` on iOS. When this option is enable, a message will be displayed to the user to turn On the bluetooth in the Settings when in airplane mode and bluetooth is Off. In our case, this message is annoying for our users.

Run the following within `AppDelegate.FinishedLaunching` before any other usage of the plugin: `BleImplementation.ShowPowerAlert(false);`.

Default value of `ShowPowerAlert` is true. The `ShowPowertAlert` will be display even if there is no options provided to `CBCentralManager` constructor.

### DIFF
--- a/Source/Plugin.BLE.iOS/BleImplementation.cs
+++ b/Source/Plugin.BLE.iOS/BleImplementation.cs
@@ -10,6 +10,7 @@ namespace Plugin.BLE
     public class BleImplementation : BleImplementationBase
     {
         private static string _restorationIdentifier;
+        private static bool _showPowerAlert = true;
 
         private CBCentralManager _centralManager;
         private IBleCentralManagerDelegate _bleCentralManagerDelegate;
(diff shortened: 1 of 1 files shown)


Thorinos:
I'm not able to see the details of the bitrise build that failed.

PR #378, opened 2019-09-04
Uwp creators update

xabre:
ToDo
- [ ] Ensure UWP device disconnect via dispose wroks
- [ ] Unify IDisposable for IDevice, IService, ICharacteristic, IDescriptor


JanNepras:
@xabre Are you still working on this pr? Could you finish it or are you ok if I finsh this one? Please let me know. We need to get this into production.

janusw:
> @xabre Are you still working on this pr?

@JanNepras I assume @xabre will not work on this anytime soon. So, feel free to take it on, if you're interested.

However, to keep things clean, I would suggest that you work on this on a new branch (probably based on the old one, or possibly rebase it to current master), and open a new PR (either directly, or once you think it's done).

> We need to get this into production.

Absolutely agree. I plan to prepare a release 2.1.2 with bug fixes and minor enhancements soon, and after that we should start planning for a 2.2.0 release that has proper support for UWP.

JanNepras:
@janusw How does it work here with versioning? What do I have to do to create preview version?

janusw:
> @janusw How does it work here with versioning? What do I have to do to create preview version?

I'm not sure if I understand exactly what your question is shooting for, so I'll answer it in two parts:
* For having a "preview" nupkg as such: I recently added a GitHub-Actions hook that builds a package for every commit (on master or in a PR). The versioning there is simply based on 'git describe'. This might already be sufficient for many people to try new versions conveniently (one can just install it locally). Btw, you might have noticed that exactly this step was failing for your last commit.
* Once we have a new "release candidate" that is somewhat tested and known to work rather well, we should probably add a tag, and I can upload the package to nuget.org (obtained the corresponding permission rights from @xabre recently).

janusw:
Btw, for my taste, the first step here should be to write down what is still left to do and where the problems are in the current UWP implementation (and put that into a new issue or PR?).

JanNepras:
@janusw I am going to fix that error in the pipeline. It seems like uwp is build successfully, but there is nothing in the log. Is it happening because it runs on mac agent? Then I am goin to create new branch and rebase. I was testing this uwp_creators_update branch and so far it works fine.

janusw:
> @janusw I am going to fix that error in the pipeline. It seems like uwp is build successfully, but there is nothing in the log. Is it happening because it runs on mac agent?

I think Plugin.BLE.UWP is not built, because the build runs on Mac, where building UWP is not supported. I guess we'll need to do that on Windows ...

janusw:
This PR is pretty outdated by now. I'm closing it in favor of #601.

PR #415, opened 2020-02-14
update docs

logeshpalani98:


janusw (commented):
I don't think the `async` keyword is needed here. Could you explain your reasons for adding it?

logeshpalani98:
Yes! sorry sir, async no need

PR #446, opened 2020-07-14
Added iOS/Mac support for 32-bit and 16-bit Service UUIDs

### LINKED ISSUE #445: iOS Adapter does not properly parse Advertisements for Service UUIDs
There is an issue in the iOS Adapter implementation of Xabre where the CBAdvertisement is not properly parsed for Service UUIDs. iOS lumps all the Service UUIDs received into a single CBAdvertisement regardless of whether or not they are 128-bit, 32-bit, or 16-bit. The UUIDs are placed in an NSArray of NSData (essentially making a jagged array). Xabre is putting all the data into a single list a then flattening it and only emits the advertisements to the upper layers as AdvertisementRecordType.UuidsComplete128Bit. So this means that the data in the advertisement records received, while using Xabre to perform the scan, could be any combination of 128-bit, 32-bit, and 16-bit service UUIDs that a device is advertising for.

## Steps to reproduce

1. Use nRFConnect to create a new Advertiser on a device. Name the device: "Test UUIDs". Set the device up to advertise its name and two 16 bit services UUIDs (1234, 2345).

2. Use the Xabre BLE.Client App on an iOS device

3. Set a break in the OnDeviceDiscovered method of the DeviceListViewModel.cs

4. Continue through each device discovery until you find the device that you set up for advertising in nrfConnect.

5. The Service UUIDs will be in a single advertisement record with 4 bytes. of data. IT should be two records, each with 2 bytes of data and a type of UuidsComple16Bitor UuidsIncomple16Bit.

## Expected behavior
Xabre needs to individually pares each NSData object in the CBAdvertisement and check its length. Then the appropriate advertisement type should be created (128, 32, 16) and forwarded to the upper App layers. That way each service UUID is received individually.

## Actual behavior
The advertisement that is received from Xabre is always a UuidsComplete128Bit with the Data being the size the total bytes from all service UUIDs actually advertised by the device.

## Configuration

**Version of the Plugin:** 2.1.1

**Platform:** iOS 13.3.1

**Device:** iPhone 8 Plus

jaredballen:
Fixes #445

### DIFF
--- a/Source/Plugin.BLE.iOS/Adapter.cs
+++ b/Source/Plugin.BLE.iOS/Adapter.cs
@@ -309,18 +309,33 @@ namespace Plugin.BLE.iOS
                     var arr = ((NSData)advertisementData.ObjectForKey(key)).ToArray();
                     records.Add(new AdvertisementRecord(AdvertisementRecordType.ManufacturerSpecificData, arr));
                 }
-                else if (key == CBAdvertisement.DataServiceUUIDsKey)
+                else if (key == CBAdvertisement.DataServiceUUIDsKey || key == CBAdvertisement.DataOverflowServiceUUIDsKey)
                 {
                     var array = (NSArray)advertisementData.ObjectForKey(key);
 
-                    var dataList = new List<NSData>();
                     for (nuint i = 0; i < array.Count; i++)
                     {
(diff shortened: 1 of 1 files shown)


xabre:
Thanks for the contribution :) Will try to release a nuget package with this fix asap.

Issue #495, opened 2021-03-21
Bluetooth 5 Advertising Extensions not supported


janusw:
Devices that advertise themselves only via Bluetooth 5 are apparently not detected by xamarin-bluetooth-le (neither with the sample app, nor with another app based on xamarin-bluetooth-le), even if the client device has hardware support for BT 5.

## Steps to reproduce

1. Set up a device that advertises only via BT 5 Advertising Extensions.

2. Build the BLE.Client.Droid app and run it (or use any other app based on xamarin-bluetooth-le).

3. Try to detect the BT5 advertisements.

## Expected behavior

The advertisements should be seen in the app.

## Actual behavior

They are not. Only BT4 (legacy) advertisements are detected.

I have used the nRF Connect app to verify that the device as such is capable of receiving the BT 5 advertisements.

## Configuration

**Version of the Plugin:** 2.1.1

**Platform:** Android 11

**Device:** Asus Zenfone 6

smsissuechecker:
Hi @janusw,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

janusw:
I think the relevant API call for Android is this one:

https://developer.android.com/reference/android/bluetooth/le/ScanSettings.Builder#setLegacy(boolean)

This is available in Android 8.0 and above, and the default value is `true` (which means that only BT 4.2 legacy advertisements are returned).

I did not find a call of this method in the xamarin-bluetooth-le source code. And in particular Plugin.BLE.Android only targets Android 7.1, where this method is not even available.

janusw:
Fixed via #496.

PR #496, opened 2021-03-22
Plugin.BLE.Android: enable BT 5 advertising extensions (#495)

janusw:
* by default only legacy advertisements (4.2 and below) are detected
* see https://developer.android.com/reference/android/bluetooth/le/ScanSettings.Builder#setLegacy(boolean)
* the target framework version needs to be increased to 8.0 for this


janusw:
Fixes #495.

PR #519, opened 2021-06-09
Fix Android autoConnect flag

### LINKED ISSUE #182: Autoconnect on Android

Hoidi:
Resolves #182 

Basically only closes the gatt instance when autoConnect was set to false when connecting to that device

### DIFF
--- a/Source/Plugin.BLE.Android/Adapter.cs
+++ b/Source/Plugin.BLE.Android/Adapter.cs
@@ -19,7 +19,6 @@ namespace Plugin.BLE.Android
 {
     public class Adapter : AdapterBase
     {
-        private ConnectParameters connectParameters;
         private readonly BluetoothManager _bluetoothManager;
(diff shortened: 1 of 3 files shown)


ske66:
Incredible. I just found this bug 30 minutes ago and there's already a fix. Brilliant work!

Hoidi:
@xabre Would you mind taking a look at this?

Hoidi:
@janusw Any opinions on this?

janusw (approved):
This essentially looks ok to me (only the `GetConnectParameters` property could be slightly simplified).

Sorry that it took so long to review this!

Khadeeja-Madathil:
After setting auto connect to true for Android devices, ConnectToKnownDeviceAsync always gives TaskCanceledException. Works fine on removing the autoconnect flag.

PR #552, opened 2021-12-01
Adds GetKnownDevicesByIds-Method to get paired devices without the need of connecting to them

MKuckert:
Such a method was missing for iOS as `GetSystemConnectedOrPairedDevices` returned only the currently connected devices. Androids `GetSystemConnectedOrPairedDevices` works just fine.

Introduces a new method to introduce no breaking change for `GetSystemConnectedOrPairedDevices` on iOS, if that method would start to return non-connected, paired devices too.


janusw:
Somehow the GHA build failed here, but the logs are not available any more. I just made some GHA fixes on master. Could you rebase this, in order to see if the problem goes away? (Also: Sorry for the late review! 🙈)

janusw (approved):
Ok, fine with me.

petetataliba:
how does this work for IOS

GetKnownDevicesByIds??

having a bit of trouble getting it to return devices.  this should return paired devices whether or not they are actively connected?  basically if i can see them in settings as bluetooth devices i would be interested to return them

PR #571, opened 2022-03-15
BLE.Client.Droid: update to Android 11

janusw:


jamesmontemagno (approved):
LGTM, i can confirm things working when targeting API 30 as well

janusw:
> LGTM, i can confirm things working when targeting API 30 as well

Thanks for the confirmation! Merged.

PR #584, opened 2022-04-11
GitHub Actions: run tests

### LINKED ISSUE #583: GitHub Actions: run unit tests
The GitHub-Actions CI build currently does the following: It builds the lib and the client app and generates nuget packages. One thing that is still missing is running the unit tests (XUnit-based).

janusw:
This extends the GHA config to build & run the unit tests. As a preparation, the test project is updated to .NET 5 and the cake script is extended to support the unit test. GHA runs the corresponding cake targets in the end. Fixes #583.

### DIFF
--- a/.build/build.cake
+++ b/.build/build.cake
@@ -83,12 +83,6 @@ Task("BuildClients")
   BuildProject("BLE.Client", "BLE.Client.macOS", Path.Combine("clients", "macOS"));
 });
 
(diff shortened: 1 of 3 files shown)


janusw:
Motivation: The Bitrise builds have been disabled recently, so the GHA builds need to be complete, since they are the only CI hook that is left now.

PR #578, opened 2022-03-28
Feature/additional scan filtering

AdamDiament:
* Adds scan filters for manufacturerid and mac address to android scanning
* Updates the sample app with a new filtered scan tab
* Added instructions to the readme

PR #588, opened 2022-04-14
Scan filtering on service data, manufacturer data or device name

janusw:
This is a follow-up to @AdamDiament's #578. It further extends the scan-filtering capabilities (on Android) with the option to filter on service data.
It also fixes a problem with the overloading of `IAdapter.StartScanningForDevicesAsync` that was introduced in #578 as well as a long-standing typo in a class name in one of the extension classes.

### DIFF
--- a/Source/BLE.Client/BLE.Client/ViewModels/DeviceListViewModel.cs
+++ b/Source/BLE.Client/BLE.Client/ViewModels/DeviceListViewModel.cs
@@ -317,16 +317,16 @@ namespace BLE.Client.ViewModels
             if (!string.IsNullOrWhiteSpace(ManufacturerIds))
             {
                 var manuIds = ManufacturerIds.Split(',');
-                var list = new List<int>();
+                var list = new List<ManufacturerDataFilter>();
                 foreach (var id in manuIds)
                 {
                     if (int.TryParse(id, out var manuId))
                     {
-                        list.Add(manuId);
+                        list.Add(new ManufacturerDataFilter(manuId));
                     }   
                 }
 
-                scanFilterOptions.ManufacturerIds = list.ToArray();
+                scanFilterOptions.ManufacturerDataFilters = list.ToArray();
             }
             if (!string.IsNullOrWhiteSpace(DeviceAddresses))
             {
--- a/Source/Plugin.BLE.Abstractions/AdapterBase.cs
+++ b/Source/Plugin.BLE.Abstractions/AdapterBase.cs
@@ -80,7 +80,7 @@ namespace Plugin.BLE.Abstractions
             }
         }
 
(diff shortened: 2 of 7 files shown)


janusw:
@AdamDiament, any comments on this?

AdamDiament:
> @AdamDiament, any comments on this?



> @AdamDiament, any comments on this?

Yes, the null default parameters on the ManufacturerData and mask data in ManufacturerDataFilter break the filter if no mask/mask data is provided. That's why I had used Array.Empty<byte>() in my version.

I have verified that changing the constructor to this fixes things

```
        public ManufacturerDataFilter(int mid, byte[] data = null, byte[] mask = null)
        {
            ManufacturerId = mid;
            ManufacturerData = data ?? Array.Empty<byte>();
            ManufacturerDataMask = mask ?? Array.Empty<byte>();
        }
```

I presume we should do the same with service Data and Mask too in ServiceDataFilter, but I cannot verify this one with testing as I don't have a device that I can set a mask for.

Other than that the code looks good.

janusw:
> Yes, the null default parameters on the ManufacturerData and mask data in ManufacturerDataFilter break the filter if no mask/mask data is provided. It acts as if no filter was added at all and just returns all devices. That's why I had used `Array.Empty<byte>()` in my version.

Good point. I also wondered about this, but did not get around to try it yet. Did it now and can confirm the issue. Will fix.

> I presume we should do the same with service Data and Mask too in ServiceDataFilter, but I cannot verify this one with testing as I don't have a device that I can set a mask for.

Tried it with `ServiceDataFilter`. It is affected by the same problem as well.

Thanks for the review!

janusw:
> I have verified that changing the constructor to this fixes things
> 
> ```
>         public ManufacturerDataFilter(int mid, byte[] data = null, byte[] mask = null)
>         {
>             ManufacturerId = mid;
>             ManufacturerData = data ?? Array.Empty<byte>();
>             ManufacturerDataMask = mask ?? Array.Empty<byte>();
>         }
> ```

In dc8bc3b806b8232d7fb288349468f1fbf8f44e2f, I am following your advice to use `Array.Empty` now, but only for the data, not the mask. If the mask is absent, I'm simply not passing it to the `ScanFilter.Builder`.

I have verified that this works well for the `ServiceDataFilter` case. @AdamDiament, could you re-check the `ManufacturerDataFilter` with this latest commit and tell me if you see any remaining issues?

janusw:
By now I have also done some more testing of the device-name filter. Seems to work well, but one thing that was slightly surprising to me was that it really only filters on advertisement records of type `CompleteLocalName` (possibly also `ShortLocalName` - could not verify that), but e.g. not on `ServiceData` records from the same device (with the same device name).

I also verified that combining a device-name and a service-data filter yields both kinds of results (i.e. both filters are OR-combined).

AdamDiament (approved):
All looks good to me

janusw:
> All looks good to me

Awesome, thanks for the feedback. Just merged it to master.

PR #594, opened 2022-05-12
Added null-check to StartScanningForDevicesNativeAsync

fozzzgate:
In Adapter.cs, if StartScanningForDevicesNativeAsync is called without specifying ScanFilterOptions (via the IAdapter overload which takes no parameters), then the lack of a null-check causes a crash.

In DeviceListViewModel.cs, I had previously submitted a pull request to add a null-check which had been accepted (commit 2317429f8a5e7153c00e090e9a66ccb3f49e4ad0), but it used the old-style syntax for null-guarding. This pull request updates it to the new C# syntax for null-guarding.


janusw (approved):
LGTM. Thanks for the contribution!

PR #595, opened 2022-05-14
Added null-checks

fozzzgate:
Added additional null-checks to prevent intermittent crashes during GATT Service queries


fozzzgate:
According to [this Stack Overflow answer](https://stackoverflow.com/questions/45347160/is-it-better-to-return-an-empty-task-or-null-c-sharp), it is perfectly acceptable to return null from an async method, so these this way of handling a null value returned from GetServicesNativeAsync should be fine.

fozzzgate:
However, the Android code for GetServicesNativeAsync() returns an empty list rather than null, so I am going to close this PR and submit a new one that returns an empty list in order to match the Android null handling.

PR #596, opened 2022-05-15
Added null-checks

fozzzgate:
Revised null handling that returns an empty list from GetServicesNativeAsync() in order to match the way that the Android code handles errors in GetServicesNativeAsync()


janusw (approved):
No objections. Thanks!

PR #598, opened 2022-05-19
Bump Microsoft.Toolkit.Uwp.Connectivity to version 6.1.1 and .NETCore to 6.2.13

fozzzgate:


janusw (approved):
Ok. Apparently you're updating as far as you can without running into the issues described in https://github.com/xabre/xamarin-bluetooth-le/issues/591#issuecomment-1125306120.

PR #599, opened 2022-05-24
Fix for Acr.UserDialogs bait library problem

fozzzgate:
UWP support in Acr.UserDialogs broke after version 7.1.0.514. And since the project is now abandoned and there is no way to open an issue, I don't know if we can get this fixed. If I create a brand-new empty UWP project and add a NuGet reference to the latest ACR.UserDialogs, I get the same bait library error. Maybe we could email or message the author, Allan Richie (@aritchie).

As a temporary fix until we can figure out the correct fix, I've rolled back the version of Acr.UserDialogs to 7.1.0.514

Visual Studio automatically removed the <ItemGroup> references to the folder 'Helpers', apparently because since the C# file in that folder was referenced directly in the project, the folder reference is superfluous. In any case, the folder still shows up in all the non-UWP projects (where it is actually used).


janusw (approved):
I can confirm that this makes BLE.Client.UWP start up without exceptions, and also scanning works well. 🥳

Great work, thank you. 👍

PR #600, opened 2022-06-02
Dont use same event handler for notifications as for read

larste-cacharge:
In our application the value received by an update/notification is not the same as the value read.

So in some cases while we´re doing a read we receive an update and both returns the notification value.

We previously used cordova-plugin-ble-central and they seems to handle reads and notifications differently.


janusw (approved):
Essentially looks good to me. No objections. Will merge soon ...

janusw:
Merged. Thanks for the contribution!

PR #606, opened 2022-06-29
Add ScanMatchMode for Android

Bodwad:
Merge request to add ScanMatchMode settings for Android BLE scan. There is no IOS equivalent.

This allows the BLE Scan to return results as soon as they are seen to speed up scans and allows for faster detection of slow frequency advertising beacons.

By default the device uses Sticky unless the user specifies Aggressive before setting the scan. There are still known issue with Samsung devices which seem to use a different.

### DIFF
--- a/Source/MvvmCross.Plugins.BLE.UWP/MvvmCross.Plugins.BLE.UWP.csproj
+++ b/Source/MvvmCross.Plugins.BLE.UWP/MvvmCross.Plugins.BLE.UWP.csproj
@@ -11,7 +11,7 @@
     <AssemblyName>MvvmCross.Plugins.BLE.UWP</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.17763.0</TargetPlatformVersion>
+    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.18362.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.16299.0</TargetPlatformMinVersion>
     <MinimumVisualStudioVersion>14</MinimumVisualStudioVersion>
     <FileAlignment>512</FileAlignment>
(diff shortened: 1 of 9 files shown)


janusw:
> This allows the BLE Scan to return results as soon as they are seen to speed up scans and allows for faster detection of slow frequency advertising beacons.

So the aggressive mode should return results faster, and possibly even detect more advertisements?


> By default the device uses Sticky unless the user specifies Aggressive before setting the scan. There are still known issue with Samsung devices which seem to use a different.

What kind of "known issue" do you refer to here? Only a different default value, or something else?

janusw (changes requested):
It would be great if you could update your first commit to get rid of the UWP changes directly (instead of reverting them in a later commit), and remove the third and fourth commit from the branch/PR. You can do this via an interactive rebase and then force-pushing the branch ....

Bodwad:
> > This allows the BLE Scan to return results as soon as they are seen to speed up scans and allows for faster detection of slow frequency advertising beacons.
> 
> So the aggressive mode should return results faster, and possibly even detect more advertisements?

Yes this is correct so the Android documentation for both sticky and Aggressive is the following

_MATCH_MODE_STICKY: For sticky mode, higher threshold of signal strength and sightings is required before reporting by hw_
_MATCH_MODE_AGGRESSIVE: In Aggressive mode, hw will determine a match sooner even with feeble signal strength and few number of sightings/match in a duration._

When set to aggressive I have also set the 'SetNumOfMatches' property on the ScanSettings to 'MATCH_NUM_ONE_ADVERTISEMENT' (Match one advertisement per filter). This means we only need to see a single advert before reporting raising the 'OnScanResult' event for the found device.

> 
> > By default the device uses Sticky unless the user specifies Aggressive before setting the scan. There are still known issue with Samsung devices which seem to use a different.
> 
> What kind of "known issue" do you refer to here? Only a different default value, or something else?

Apologies, I didn't seem to finish the comment. On Samsung devices it doesn't appear to take these setting verbatim. It appears to improve the scan frequency of adverts but it doesn't seem to take the 'MATCH_NUM_ONE_ADVERTISEMENT' setting correctly. When testing on a Pixel 4 and on one of our own custom android devices we see adverts coming in one for one from the advertising peripheral. On a Samsung device (XCover) we don't seem to get an OnScanResult for each advert transmitted by the peripheral, we get a reduced number which may indicate that the SetNumOfMatches command was ignored.

JayMansel:
Just wanted to reference https://github.com/xabre/xamarin-bluetooth-le/issues/136#issuecomment-271848590 which also requests this functionality

Bodwad:
> Thanks for the comments, @Bodwad. This is certainly a useful addition to the API, even if it is Android-only.
> 
> However, your two newly-added commits are doing not quite what I was requesting. (Rebasing is sort of an advanced git feature that not everyone is comfortable with, I suppose.) I can take care of this myself, but I'll probably need to open a new PR that replaces this one.

Yeah sorry I have never done this before. I thought I had done it correctly but it seems to have just added more commits :(

janusw:
> Yeah sorry I have never done this before. I thought I had done it correctly but it seems to have just added more commits :(

Don't worry. I opened #607 now, so this one should be closed.

PR #607, opened 2022-06-30
Added Scan Match Mode to Android

janusw:
This is a cleaned-up version of #606, where everything is squashed into one commit (and all unnecessary parts are removed).

@Bodwad, could you please double-check this, to make sure it is still working as intended and I did not mess up anything?

### DIFF
--- a/Source/Plugin.BLE.Android/Adapter.cs
+++ b/Source/Plugin.BLE.Android/Adapter.cs
@@ -149,6 +149,17 @@ namespace Plugin.BLE.Android
 
             var ssb = new ScanSettings.Builder();
             ssb.SetScanMode(ScanMode.ToNative());
+            ssb.SetMatchMode(ScanMatchMode.ToNative());
+
+            // If set to agressive reduce the number of adverts needed before raising the DeviceFound callback
+            if (ScanMatchMode.ToNative() == BluetoothScanMatchMode.Aggressive)
+            {
+                // Be more agressive when seeking adverts
+                ssb.SetNumOfMatches((int)BluetoothScanMatchNumber.OneAdvertisement);
+                ssb.SetCallbackType(ScanCallbackType.AllMatches);
+                Trace.Message("Using ScanMatchMode Agressive");
+            }
+            
             if (Build.VERSION.SdkInt >= BuildVersionCodes.O)
             {
                 // enable Bluetooth 5 Advertisement Extensions on Android 8.0 and above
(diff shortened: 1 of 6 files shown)


janusw:
Btw, there was an API-level check missing for backwards compatibility. The match mode was introduced in API level 23:

https://developer.android.com/reference/android/bluetooth/le/ScanSettings.Builder#setMatchMode(int)

I added this check in b565ddcd19cca5f141217032dc425e6075fa63a2.

Bodwad:
> This is a cleaned-up version of #606, where everything is squashed into one commit (and all unnecessary parts are removed).
> 
> @Bodwad, could you please double-check this, to make sure it is still working as intended and I did not mess up anything?

Yes I can confirm I have checked and this is all working as expected

Source/Plugin.BLE.Android/Adapter.cs
```diff
@@ -149,6 +149,22 @@ private void StartScanningNew(ScanFilterOptions scanFilterOptions)
...
+            {
+                // set the match mode on Android 6 and above
+                ssb.SetMatchMode(ScanMatchMode.ToNative());
+
+                // If set to agressive, reduce the number of adverts needed before raising the DeviceFound callback
+                if (ScanMatchMode.ToNative() == BluetoothScanMatchMode.Aggressive)
+                {
+                    // Be more agressive when seeking adverts
+                    ssb.SetNumOfMatches((int)BluetoothScanMatchNumber.OneAdvertisement);
+                    ssb.SetCallbackType(ScanCallbackType.AllMatches);
```
janusw, line 163: @Bodwad  This call to `SetCallbackType` is not really necessary, since `AllMatches` is the default, right?

Bodwad: > @Bodwad This call to `SetCallbackType` is not really necessary, since `AllMatches` is the default, right?

I was trying to work this out as I had noticed the commented out line that already existed in Master, I presume this is why it got commented out. 

There doesn't seem to be anything in the Android documentation about default values. As a result I would personally specify it as we know that is definitely what we want in this scenario. It would also override any potential 'Manufacturer OS' defaults that may exist.

If you think it isn't needed then feel free to remove it

janusw: `AllMatches` indeed seems to be the default, and it should really be used for all scenarios (not just the aggressive mode).

It doesn't matter much if we set it explicitly, but code-wise it looks weird if it's only set for aggressive. So I'd rather not set it at all (will remove that line).

janusw:
I have merged this now. Thanks for the contribution, @Bodwad!

Bodwad:
Brilliant thank you, thanks for your help

Bodwad:
Is there a planned release date for these changes? Either to pre-release or official release would be great as we need to update a few of our projects that use this official nugget package :)

janusw:
> Is there a planned release date for these changes?

I planned to do another pre-release. Will tag and package it this week ...

janusw:
Release 2.2.0-pre5 has been published last night!

PR #614, opened 2022-08-11
Upgrade to .NET 6

jamesmontemagno:
So, this is a big one, and really simplifies the structure of the project into a modern .NET project and will make it way easier to package, put into CI/CD, and so much more. Let me know your thoughts. The only odd thing that I saw was in IBleCentralManagerDelegate as CBPeripheralsEventArgs no longer exists, but it doesn't look like we use it at all.

This will maintain backwards compat with xamarin.ios/android as well as uwp


jamesmontemagno:
So, this is a big one, and really simplifies the structure of the project into a modern .NET project and will make it way easier to package, put into CI/CD, and so much more. Let me know your thoughts. The only odd thing that I saw was in IBleCentralManagerDelegate as CBPeripheralsEventArgs no longer exists, but it doesn't look like we use it at all.

janusw (changes requested):
@jamesmontemagno That's awesome, thanks for looking into this!

Note that the GitHub Actions build currently fails with this error (in the `Restore` step):
```
Restoring D:/a/xamarin-bluetooth-le/xamarin-bluetooth-le/Source/BLE.sln
MSBuild auto-detection: using msbuild version '17.2.1.25201' from 'C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin'.
Error: C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\6.0.400\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.ImportWorkloads.targets(38,5): error NETSDK1147: To build this project, the following workloads must be installed: android [D:\a\xamarin-bluetooth-le\xamarin-bluetooth-le\Source\Plugin.BLE\Plugin.BLE.csproj]
Error: C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\6.0.400\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.ImportWorkloads.targets(38,5): error NETSDK1147: To install these workloads, run the following command: dotnet workload restore [D:\a\xamarin-bluetooth-le\xamarin-bluetooth-le\Source\Plugin.BLE\Plugin.BLE.csproj]
```

The GHA config already uses .NET 6, only the workloads need to be installed apparently (android, ios, etc). `dotnet workload restore` should be sufficient for  this, it seems.

Could you please update [.github/workflows/dotnet.yml](https://github.com/xabre/xamarin-bluetooth-le/tree/master/.github/workflows/dotnet.yml) accordingly?

jamesmontemagno:
On it!

jamesmontemagno:
So... i think i have updated it to install the .NET MAUI workload appropriately here... I also do the pack directly in the github action and not in the cake script as it is easier to just be in there. Let's see how that works out, but may need a bit of help here

janusw:
> So... i think i have updated it to install the .NET MAUI workload appropriately here... I also do the pack directly in the github action and not in the cake script as it is easier to just be in there.

In principle that should be fine (although we separated those steps in GHA in order to have more clarity about occurring errors, as in "restore and build works, only packaging fails" etc). Matter of taste, I guess.

However, it seems that your new build step comes too early in dotnet.yml (in particular: before the `setup-msbuild` step) and therefore fails with the error:
`The term 'msbuild' is not recognized as a name of a cmdlet, function, script file, or executable program.`

.github/workflows/dotnet.yml
```diff
@@ -7,6 +7,9 @@ on:
     branches: [ '**' ]
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
+  
+env:
+    NUGET_VERSION: 3.0.0
```
janusw, line 12: Two short comments about this version number:
* In principle we're working towards a version 2.2.0 right now. So you are suggesting to bump it directly to 3.0 with the .NET 6 changes?
* Currently we use the output of `git describe` for versioning the nupkgs built via GHA. This has the advantage that it gives better distinction of off-release commits on master, and versioning is essentially controlled via tags (i.e. we get a version 3.0 as soon as we set the corresponding tag).

.github/workflows/dotnet.yml
```diff
@@ -58,6 +69,7 @@ jobs:
       run: nuget pack .build/MvvmCross.Plugin.BLE.nuspec -Version $(git describe)
     - name: Upload packages
       uses: actions/upload-artifact@v2
+      if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
```
janusw, line 72: So this only uploads the artifacts for pushes to master? Any particular reason why you apply that restriction? nupkgs from a PR can be quite useful for people wanting to try a branch with an external project that references Plugin.BLE. I've done this myself a number of times.

jamesmontemagno: Let me remove i didn't know what was the right way. I use devops for everythign so my system is very different.

.github/workflows/dotnet.yml
jamesmontemagno: i would put out 2.2.0 first and then 3.0 as a beta probably

jamesmontemagno:
let's see if that fixes it up

janusw:
> let's see if that fixes it up

Much better already, but now GHA fails with:

```
   2 |  … E\Plugin.BLE.csproj /p:Configuration=Release /t:restore;build;pack /p …
     |                                                            ~~~~~
     | The term 'build' is not recognized as a name of a cmdlet, function, script file, or executable
     | program. Check the spelling of the name, or if a path was included, verify that the path is correct
     | and try again.
```

jamesmontemagno:
INTERESTING..... alright i'll check it out

janusw:
> Much better already, but now GHA fails with:
> 
> ```
>    2 |  … E\Plugin.BLE.csproj /p:Configuration=Release /t:restore;build;pack /p …
>      |                                                            ~~~~~
>      | The term 'build' is not recognized as a name of a cmdlet, function, script file, or executable
>      | program. Check the spelling of the name, or if a path was included, verify that the path is correct
>      | and try again.
> ```

Sooo ... this is apparently some weird msbuild bug. I can reproduce it with msbuild 17.3.1 on Windows 10.

The [documentation ](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference) says:
```
Specify each target separately, or use a semicolon or comma to separate multiple targets
```
But in practice only commas works, while semicolons produce the error above. Ouch 🙈

jamesmontemagno:
alright, let's give it a go

janusw:
Ok, we're getting there (slowly) 😉.

The next error now is:
```
Error: D:\a\xamarin-bluetooth-le\xamarin-bluetooth-le\Source\Plugin.BLE\obj\Release\netstandard2.0\Plugin.BLE.AssemblyInfo.cs(23,55):
error CS7034: The specified version string does not conform to the required format - major[.minor[.build[.revision]]]
[D:\a\xamarin-bluetooth-le\xamarin-bluetooth-le\Source\Plugin.BLE\Plugin.BLE.csproj]
```

Seems like your `NUGET_VERSION: 3.0.0-beta` is not fit to be used as an AssemblyVersion?

jamesmontemagno:
WEIRD! let me just do 3.0.0.0 lol

jamesmontemagno:
It is odd because i am using the same command and steps in ADO, so odd.

jamesmontemagno:
ok, so that should totally do it, i actually did it wrong. So i now build 2 packages... beta and non-beta so you can decide which to ship basically.

jamesmontemagno:
let me fix up the tests lol

jamesmontemagno:
Ok, should be good and ran tests :)

pm64:
Just a side thought here.. once this update is completed, should we contemplate renaming the library to "maui-bluetooth-le"?

janusw:
> Just a side thought here.. once this update is completed, should we contemplate renaming the library to "maui-bluetooth-le"?

Renaming might be reasonable, but since @jamesmontemagno's changes are backwards compatible with Xamarin, it's probably not the best idea to use "MAUI" in the name. Maybe something more general like "Plugin.BLE" (the current name of the nupkg) or "BLE.NET" / "BLEDotNet" ...?

pm64:
I like all of those suggestions, but I'm not familiar enough with this project to know which is most appropriate. I'm in favor of anything that drops "Xamarin" from the name. As close as Xamarin is to my heart, it now implies legacy software. I think? Maybe @jamesmontemagno can offer an insider opinion.

beeradmoore:
Not related specifically to this project and don't want to take it off topic, but I have been thinking lately what is a standardised naming convention we should be using.

I have Xam.Plugin.Mixpanel.iOS,  Xam.Plugin.Mixpanel.Android and Xam.Plugin.LinkedIn.LiTr.

I was considering Net.iOS.Mixpanel, but it's a little gross. So keeping an eye on this to see what scheme people come up with.

jamesmontemagno:
@janusw let's see if that works. would be faster if i could run the workflows i think...


On the packaging of names.. Plugin.BLE seems ust fine. I think probably just re-naming the github repo to "dotnet-bluetooth-plugin" would work. I always prefer dotnet over "net" or DotNet over Net for a namespace

janusw:
> @janusw let's see if that works. would be faster if i could run the workflows i think...

Yeah, since you have no commits on master (yet), you are classified as a first-time contributor and your workflow runs need to be approved manually.

jamesmontemagno:
WE ARE GREEN!

naryk86:
Will the changes resolve also problems with "Windows Machine" ?
([Issue 421](https://github.com/xabre/xamarin-bluetooth-le/issues/421)

jamesmontemagno:
@janusw anything left for this?

janusw:
> @janusw anything left for this?

Hi @jamesmontemagno, sorry for the late feedback here (and thanks again for the great contribution). This looks mostly ok to me, but it seems you're only converting the Plugin.BLE.* projects to .NET 6, but not MvvmCross.Plugins.* or the BLE.Client.* samples. Could you comment on this? Would you be willing to do the remaining conversions as a follow-up? Are there any obstacles for these cases?

jamesmontemagno:
I could.... I don't know much about the in general and if MVVMCross is updated to .NET 6? I think it is... are these packages really needed though? I would love to get a beta out as I can't update my apps yet due to this not being updated.

As for the samples, they will continue to work as is for Xamarin samples since we still target them. We could work on new .NET MAUI samples as needed.

jamesmontemagno:
Alright, convert it over to .NET 6... i am not sure how the dependency stuff works for building multiple nuget packages together though.....

jamesmontemagno:
@janusw can you give me access to run the github actions please. this going back and forth takes a long time when I can just run and validate on my side.

jamesmontemagno:
I think this failure is because of some odd build stuff with VS being installed and doesn't have the xamarin bits in it... hmmmm

janusw:
> @janusw can you give me access to run the github actions please.

I'd love to, but unfortunately I don't know how to do this. Do you?

Note that I'm not the owner of this repo and don't have full admin permissions. Possibly @xabre needs to help out here ...?

jamesmontemagno:
try again now, put in workaround for breaking changes

twenzel:
@xabre would you promote @janusw to repository admin in order to get some github actions updated?

jamesmontemagno:
Should be good now :)

janusw:
> Should be good now :)

Looks great, we're green again. Thanks a lot! (Please give me a few more days to review and test this.)

Btw, I have proposed to @xabre to rename the project (to `dotnet-bluetooth-le`) and move it into a new org, in order to better share maintainership. I hope he will accept.

janusw:
So, short review: All the actual conversion of the project files and code looks very solid to me. It only misses the sample app (which is not essential, and can be taken care of later).

One small remaining issue that I'd like to see fixed, though, is the fact that the GHA build currently does not show any artifacts. Probably just the path / filename in the upload step does not match? Could you please fix that, @jamesmontemagno? Having artifacts makes it much easier for people to try the results of this PR.

And then I have some more minor points of criticism concerning the build / CI / Github Actions. I had some comments on this before, and I'm still not fully happy with the changes in this area:
* I don't like the fact that there are two builds of each nupkg  in GHA, which are exactly the same, except that one is named "*-beta".
* I don't like the fact that the versioning is fixed to 3.0.0 (instead of the `git describe` we had before). I think it is useful to have a unique versioning code for each CI build. `git describe` was our way of achieving that, but a different mechanism might also be acceptable.
* I don't like the fact that you removed the possibility to build the libs via the cake script. Building without cake in GHA is ok (it's sufficiently simple), but it would be nice to keep the option to build via cake (if possible).

I can possibly just take care of most of this in a follow-up PR (unless you want to fix some of it yourself already) ...

janusw (changes requested):
https://github.com/xabre/xamarin-bluetooth-le/pull/614#issuecomment-1294703666

janusw:
> Btw, I have proposed to @xabre to rename the project (to `dotnet-bluetooth-le`) and move it into a new org, in order to better share maintainership. I hope he will accept.

Both renaming and moving has been done by now. Many thanks to @xabre!

janusw:
> One small remaining issue that I'd like to see fixed, though, is the fact that the GHA build currently does not show any artifacts. Probably just the path / filename in the upload step does not match? Could you please fix that, @jamesmontemagno? Having artifacts makes it much easier for people to try the results of this PR.

I have fixed this by now (see https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/pull/614/commits/6e9ccf712ba1c6b9c4eb17af62e8c7ef8cf04b5c). GitHub actions shows a zipped artifacts file now, which contains four nupkgs (two for Plugin.BLE and two for MvvmCross.Plugin.BLE).

I will try the vanilla package and verify that it still works well in a Xamarin context. If anyone else wants to give it a try: Feedback is certainly welcome here. In particular it would be nice to know whether the package works in a MAUI context, and whether the MvvmCross plugin works. If you try it, please let us know what you did and whether you found any issues.

RobertLinak:
> > One small remaining issue that I'd like to see fixed, though, is the fact that the GHA build currently does not show any artifacts. Probably just the path / filename in the upload step does not match? Could you please fix that, @jamesmontemagno? Having artifacts makes it much easier for people to try the results of this PR.
> 
> I have fixed this by now (see [6e9ccf7](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/commit/6e9ccf712ba1c6b9c4eb17af62e8c7ef8cf04b5c)). GitHub actions shows a zipped artifacts file now, which contains four nupkgs (two for Plugin.BLE and two for MvvmCross.Plugin.BLE).
> 
> I will try the vanilla package and verify that it still works well in a Xamarin context. If anyone else wants to give it a try: Feedback is certainly welcome here. In particular it would be nice to know whether the package works in a MAUI context, and whether the MvvmCross plugin works. If you try it, please let us know what you did and whether you found any issues.

I have tried installing the vanilla nupkg, from the zipped artifact, in a MAUI project and I am still getting this exception when accessing CrossBluetoothLE.Current on iOS: 
"This functionality is not implemented in the portable version of this assembly.  You should reference the NuGet package from your main application project in order to reference the platform-specific implementation."
On Android it works fine, which is exactly the same behavior as with the old version.

Regarding MvvmCross there is no MAUI compatible nuget available and as I understand it, the one guy maintaining it has no interest in MAUI.

RobertLinak:
Also on android I have noticed that when doing a series of sequential reads by calling await characteristic.WriteAsync(buffer); the plugin will randomly stop invoking the characteristic.ValueUpdated event, which is a breaking bug that makes the plugin no longer usable.

janusw:
> I have tried installing the vanilla nupkg, from the zipped artifact, in a MAUI project and I am still getting this exception when accessing CrossBluetoothLE.Current on iOS: "This functionality is not implemented in the portable version of this assembly. You should reference the NuGet package from your main application project in order to reference the platform-specific implementation." On Android it works fine, which is exactly the same behavior as with the old version.

Thanks for the feedback, @RobertLinak. @jamesmontemagno, do you have any comments on this? Any idea why this occurs?

janusw:
> Also on android I have noticed that when doing a series of sequential reads by calling await characteristic.WriteAsync(buffer); the plugin will randomly stop invoking the characteristic.ValueUpdated event, which is a breaking bug that makes the plugin no longer usable. Downgrading to version 2.1.3 fixes this issue.

Does this problem also occur with 2.2.0-pre5, or only on this PR?

jamesmontemagno:
I am pretty heads down on .NET Conf right now. Thanks for fixing up the github action.

I agree with everything you said on this. This package will also continue to work with Xamarin as well which is nice.

For the android stuff.. I mean no code changes to be honest so..... unless that was broken in 2.2.0?

RobertLinak:
> Does this problem also occur with 2.2.0-pre5, or only on this PR?

I would like to test it but the command Install-Package Plugin.BLE -Pre now installs version 3.0.0 so I dont know how to install 2.2.0-pre5 anymore.

janusw:
> > Does this problem also occur with 2.2.0-pre5, or only on this PR?
> 
> I would like to test it but the command Install-Package Plugin.BLE -Pre now installs version 3.0.0 so I dont know how to install 2.2.0-pre5 anymore.

`Install-Package` can take a `-Version` parameter, see https://learn.microsoft.com/en-us/nuget/reference/ps-reference/ps-ref-install-package, or you can manage packages with the VS UI (where you can specify the exact version as well).

It would be really great if you could check this. If the problem does not occur with 2.2 pre, then this would be the last major blocker here IMHO (but I rather suspect it might be introduced somewhere along the 2.2 changes).

janusw:
> Also on android I have noticed that when doing a series of sequential reads by calling await characteristic.WriteAsync(buffer); the plugin will randomly stop invoking the characteristic.ValueUpdated event, which is a breaking bug that makes the plugin no longer usable. Downgrading to version 2.1.3 fixes this issue.

In fact this sounds a bit like it might be related to PR #600. If so, then 2.2.0-pre4 should still be working, but 2.2.0-pre5 should fail. If this is the case, please open a new issue for it.

RobertLinak:
The VS UI does not show any versions beyond 2.1.3, but I tried the different -pre versions using the -Version parameter.

2.2.0-pre1 works
2.2.0.pre2 introduces the breaking bug
There is no difference between pre4 and pre5 in my test.

I can open a new issue if you want but I am not really able to investigate or work on this project as I have no idea how you actually compile (VS shows a lot of errors for me) or pack all the projects into a single nuget.

RobertLinak:
If the purpose of this #614 is to make the plugin compatible with MAUI it would be nice to get some clarification or a sample project that shows how to actually setup and use it for iOS in MAUI, since going with the current info in the readme it only works for android.

janusw:
> The VS UI does not show any versions beyond 2.1.3, but I tried the different -pre versions using the -Version parameter.
> 
> 2.2.0-pre1 works 2.2.0.pre2 introduces the breaking bug There is no difference between pre4 and pre5 in my test.
> 
> I can open a new issue if you want

Yes, please open a new issue. This is very valuable information, and it shows that the issue is not related to this PR, but was present before. Thanks!

janusw:
> If the purpose of this #614 is to make the plugin compatible with MAUI it would be nice to get some clarification or a sample project that shows how to actually setup and use it for iOS in MAUI, since going with the current info in the readme it only works for android (which it already did in version 2.1.3).

AFAIU the main intention of this PR is to update the project structure to .NET6 style, which I assume is a prerequisite for being compatible with MAUI, but possibly does not go all the way there? @jamesmontemagno, could you comment on this? Have you tried the package with a MAUI app yet, and have you encountered the same issues on iOS?

At this point I can confirm that the nupkg from this PR appears to work well in a Xamarin context (tested on Android, iOS, Mac). That is, for my purposes at least (which do not cover the whole API).

twenzel:
The 2.2.0-pre5 version is already functional with .NET MAUI.

RobertLinak:
> The 2.2.0-pre5 version is already functional with .NET MAUI.

Do you mean that you are able to access CrossBluetoothLE.Current on anything other than android (like iOS) without catching an exception?
I agree that it can be installed in a MAUI project, but we are missing some clarification on usage for iOS as an example would be nice.

twenzel:
> > The 2.2.0-pre5 version is already functional with .NET MAUI.
> 
> Do you mean that you are able to access CrossBluetoothLE.Current on anything other than android (like iOS) without catching an exception? I agree that it can be installed in a MAUI project, but we are missing some clarification on usage for iOS. If anyone has a working iOS sample in a MAUI project then please upload it to be used as guidance for others.

You're right. I've a working solution on Android. I can't test on iOS yet because VS 2022 does not support XCode 14.1 yet..

ComtelJeremy:
I am able to build on iOS, to test. I'm stuck on this issue, right now and have a working Android app, but not iOS.
However (dumb question), what's the best way to pull in this PR to my MAUI project to test a build?

I believe the Exception we are getting is due to the references in the  being for "xamarin", not "ios" and therefore the proper assembly isn't pulled in from the nuget. Am I correct?

RobertLinak:
> You're right. I've a working solution on Android. I can't test on iOS yet because VS 2022 does not support XCode 14.1 yet..
The newest update for VS 2022 Preview supports building through XCode 14.1 (14B47b)

RobertLinak:
> I am able to build on iOS, to test. I'm stuck on this issue, right now and have a working Android app, but not iOS. However (dumb question), what's the best way to pull in this PR to my MAUI project to test a build?
> 
> I believe the Exception we are getting is due to the references in the being for "xamarin", not "ios" and therefore the proper assembly isn't pulled in from the nuget. Am I correct?

You can download the nuget from the artifacts section at the bottom of this page
https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/actions/runs/3376811300
And then in your VS nuget package manager theres a little cogwheel in the top right corner where you can set Package source to a local path you define.

RobertLinak:
On line 32 of CrossBluetoothLE.cs there is a compiler condition #if NETSTANDARD which is entered when building on iOS, which then causes the BleImplementation to not be instantiated and the NotImplementedInReferenceAssembly exception to be thrown.
But why iOS is being handled by the compiler as NETSTANDARD i dont understand.
When choosing the compiler option net6.0-ios from the dropdown then VS shows that the code instantiating BleImplementation should be run.

janusw:
> And then I have some more minor points of criticism concerning the build / CI / Github Actions. I had some comments on this before, and I'm still not fully happy with the changes in this area:
> 
>     * I don't like the fact that there are two builds of each nupkg  in GHA, which are exactly the same, except that one is named "*-beta".
> 
>     * I don't like the fact that the versioning is fixed to 3.0.0 (instead of the `git describe` we had before). I think it is useful to have a unique versioning code for each CI build. `git describe` was our way of achieving that, but a different mechanism might also be acceptable.

Reply to self: Both of these points should be fixed with 18dc86226eb613355708e91c50d00826a8780ca0, which introduces `git describe` for CI-versioning again. I think this makes much more sense, because
a) it makes it easier to distinguish different CI-generated nupkgs (each has its own unique version) and
b) it avoids the danger of confusing these packages (which are really just random builds on a random branch by a random dude 😉) with an actually-final 3.0.0 package.

@jamesmontemagno, I hope these changes are ok with ok (as well as me messing with your PR 😉).


>     * I don't like the fact that you removed the possibility to build the libs via the cake script. Building without cake in GHA is ok (it's sufficiently simple), but it would be nice to keep the option to build via cake (if possible).

This is still open. Not terribly important, though. I might look into it soon (if I find the time).

janusw:
> On line 32 of CrossBluetoothLE.cs there is a compiler condition #if NETSTANDARD which is entered when building on iOS, which then causes the BleImplementation to not be instantiated and the NotImplementedInReferenceAssembly exception to be thrown.

Interesting. In fact this condition was modified by @jamesmontemagno here: https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/pull/614/commits/443a51ebcecd78febf763824b7bfea723e222552#diff-86f7fc3deec5b5432eff7dcc6bf407c47a5885078a4b4a6ed1f9e63c8540595d

Maybe one can just remove the condition fully with .NET6? Or it needs to be fixed in some other way. @jamesmontemagno, do you have any ideas on this?

janusw (approved):
> > ```
> > * I don't like the fact that you removed the possibility to build the libs via the cake script. Building without cake in GHA is ok (it's sufficiently simple), but it would be nice to keep the option to build via cake (if possible).
> > ```
> 
> This is still open. Not terribly important, though. I might look into it soon (if I find the time).

I have fixed this in https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/pull/614/commits/32a03eb50f02766ace9a432aa178f56a00444a6d. With that, all my comments have been taken care of, so I'm essentially happy with this PR. @jamesmontemagno, if you are happy with the current state as well, I'll go ahead and merge it. 🥳

janusw:
> On line 32 of CrossBluetoothLE.cs there is a compiler condition #if NETSTANDARD which is entered when building on iOS, which then causes the BleImplementation to not be instantiated and the NotImplementedInReferenceAssembly exception to be thrown.

So, I took a stab at fixing this: https://github.com/janusw/dotnet-bluetooth-le/commit/817a70c865fa6b96a57f637d2124f64e7341ef2e. @RobertLinak could you maybe take this for a test ride (nupkg [here](https://github.com/janusw/dotnet-bluetooth-le/suites/9155108191/artifacts/426046159)) and check if you still get the same exception on .NET6/iOS?


> But why iOS is being handled by the compiler as NETSTANDARD i dont understand. When choosing the compiler option net6.0-ios from the dropdown then VS shows that the code instantiating BleImplementation should be run.
> 
> To me this looks like a framework problem with MAUI and not an actual problem with this plugin.

Agree, this seems strange. Maybe the above is a suitable workaround for the time being?

RobertLinak:
I tried the new nuget 2.2.0-pre5-23-g817a70c, and it still catches the exception on iOS.

janusw:
> I tried the new nuget 2.2.0-pre5-23-g817a70c, and it still catches the exception on iOS.

Hm, too bad, but thanks for checking.

I suggest we merge this PR as is and deal with the iOS trouble in a follow-up. @jamesmontemagno, you ok with that?

RobertLinak:
So I did some testing of preprocessor directives...

When building on android it hits conditions for `__MOBILE__`, `__ANDROID__`, `NET`, `NET6_0`, `NETCOREAPP`, `ANDROID`, `NET6_0_OR_GREATER`

When building on iOS it hits pretty much nothing except for `NETSTANDARD`. It does not even hit relevant directives such as `__IOS__` or `IOS`.

If this is not related to MAUI then at least it must be some runtime or SDK related stuff. I dont see how this should be possible to solve by any normal random developer.

RobertLinak:
I tried converting the plugin.BLE project to a '.NET MAUI Class Library' project and added it as a project reference in my client solution instead of using the nuget, and I can confirm that it hits the IOS preprocessor directive and is able to scan and find devices on iOS. I double checked that the project type was the issue by once again removing the project reference and reinstalling the nuget 2.2.0-pre5-23-g817a70c and confirming that the nuget does not work.

janusw:
> I tried converting the plugin.BLE project to a '.NET MAUI Class Library' project and added it as a project reference in my client solution instead of using the nuget, and I can confirm that it hits the IOS preprocessor directive and is able to scan and find devices on iOS.

Sounds good, can you push this somewhere, where one can have a look and test it? (Just fork this repo, if you haven't done yet, and create a branch based on this one with your modifications. Btw I pushed the branch of this PR now to https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/tree/dotnet6)

> This workaround of course removes the backwards compatibility with Xamarin.

That is a problem, since we certainly want to keep Xamarin compatibility.

Maybe this whole thing is a framework bug in .NET6 that affects multi-targeting projects with Xamarin / .NET Standard compatibility?!?

RobertLinak:
I forked it here (sorry, this is the extent of my current github skills):
https://github.com/RobertLinak/dotnet-bluetooth-le
But I dont think it will be much help to you since the big problem is the preprocessor directives not being hit.
Other than converting to a .NET MAUI Class Library the only thing I did was change the folder structure and make the BleImplementation classes partial which is the only way I know how to make them platform specific in MAUI.

But maybe it can be useful for someone else who, like me, would like something that is able to build on both iOS and Android while we wait for an official fix to the plugin.

janusw:
> I forked it here (sorry, this is the extent of my current github skills): https://github.com/RobertLinak/dotnet-bluetooth-le [...] Other than converting to a .NET MAUI Class Library the only thing I did was change the folder structure and make the BleImplementation classes partial which is the only way I know how to make them platform specific in MAUI.

Thanks! Seems like you started from master. I thought you could start from the dotnet6 branch (which includes all of @jamesmontemagno's work), so that you don't have to redo the bulk of the conversion.
Anyway, looking at your [pluginBleMaui.csproj](https://github.com/RobertLinak/dotnet-bluetooth-le/blob/master/pluginBleMaui/pluginBleMaui.csproj), it seems like the major difference is that you are using net6-targets only, while @jamesmontemagno is also including Xamarin/.NET Standard for backwards compatibility. We should try to verify that this is actually what's making the difference.

> But I dont think it will be much help to you since the big problem is the preprocessor directives not being hit.

I don't understand this comment. I thought your approach _fixes_ the problem with the preprocessor symbols?!?

RobertLinak:
> I don't understand this comment. I thought your approach _fixes_ the problem with the preprocessor symbols?!?

My approach was to convert to a .NET MAUI Class Library, which fixes the preprocessor symbol issue, but I'm not sure you will be able to do that since I dont think it works with Xamarin anymore.

janusw:
> > I don't understand this comment. I thought your approach _fixes_ the problem with the preprocessor symbols?!?
> 
> My approach was to convert to a .NET MAUI Class Library, which fixes the preprocessor symbol issue, but I'm not sure you will be able to do that since I dont think it works with Xamarin anymore. But I dont know. Probably james knows how to approach it. If #614 is already using a .NET MAUI Class Library then I dont know why the preprocessor symbols dont work, since that is pretty much my only change.

Yes, AFAICS James is already using a "MAUI class library", but with multi-targeting that includes the old Xamarin targets as well, so I assume the latter is what introduces the preprocessor issues with iOS (but that is only an assumption so far).

If that's the case, then maybe we can use two separate project files for MAUI/.NET 6 and Xamarin/.NET Standard as a temporary workaround (with the same code contents)? I'm not sure if that will work, though.

RobertLinak:
> Yes, AFAICS James is already using a "MAUI class library", but with multi-targeting that includes the old Xamarin targets as well, so I assume the latter is what introduces the preprocessor issues with iOS (but that is only an assumption so far).

Ok, I tried testing with #614 and just changing the csproj to only have these targets:
```
<TargetFrameworks>net6.0-android;net6.0-ios;net6.0-maccatalyst;</TargetFrameworks>
<TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);uap10.0.19041;net6.0-windows10.0.19041;</TargetFrameworks>
```
Which fixes the preprocessor symbol issue and it is able to hit the IOS symbol.

janusw:
> > Yes, AFAICS James is already using a "MAUI class library", but with multi-targeting that includes the old Xamarin targets as well, so I assume the latter is what introduces the preprocessor issues with iOS (but that is only an assumption so far).
> 
> Ok, I tried testing with #614 and just changing the csproj to only have these targets:
> 
> ```
> <TargetFrameworks>net6.0-android;net6.0-ios;net6.0-maccatalyst;</TargetFrameworks>
> <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);uap10.0.19041;net6.0-windows10.0.19041;</TargetFrameworks>
> ```
> 
> Which fixes the preprocessor symbol issue and it is able to hit the IOS symbol.

Awesome, thanks for checking. That confirms my assumption. Sounds a lot like a framework/tooling bug to me. Would probably a be a good idea to file an issue at https://github.com/dotnet/core ...

axa88:
So... now that .NET 7 is out, is there any plan to release this?

janusw:
> So... now that .NET 7 is out, is there any plan to release this?

Well, yeah. In principle I was hoping for some feedback from @jamesmontemagno, but I guess I'll jut go ahead and merge the current state to master and package a new pre-release version.

We still need to figure out how to deal with the MAUI-iOS issue, but at least the Xamarin compatibility works well already. And maybe just updating to .NET 7 will already fix the iOS issues (and should be rather simple) ...

janusw:
Merged to master. Thanks again for this valuable contribution, @jamesmontemagno!

axa88:
I too appreciate all the effort...

> And maybe just updating to .NET 7 will already fix the iOS issues (and should be rather simple) ...

About that... I'm just a novice end user really but, what would this take?
I tried tuning things but couldn't get a cross platform project to build using latest .net 7 release.

Will be watching for updates.

janusw:
> > And maybe just updating to .NET 7 will already fix the iOS issues (and should be rather simple) ...
> 
> About that... I'm just a novice end user really but, what would this take?

Should be much simpler than the transition to .NET6. Ideally it would be enough to simply update the target frameworks from `net6.0-*` to `net7.0-*`, but I'm not sure if this is really sufficient.

twenzel:
Please use multi-targeting to support both net6 and net7.

jamesmontemagno:
There isn't a need to really swap to .net 7 for compilation unless we are using newer features in ios.

net6 should be good to go, a .net 7 project will pull in .net 6 just fine.... although I am fine compiling against net7 exclusively anyways as any .net maui should upgrade regardless in my opinion.

as for the ios issue... i havent' looked into it yet, but should be able to this week or next week sometime... i mean should just be able to pull up the nuget and see. I downloaded it and it was all looking good in the nuget package explorer

janusw:
> There isn't a need to really swap to .net 7 for compilation unless we are using newer features in ios.
> 
> net6 should be good to go, a .net 7 project will pull in .net 6 just fine.... 

In principle I agree, but if the iOS issue with the preprocessor symbols is actually an SDK/tooling issue, it might potentially be fixed in v7, I guess. That might be a reason to swap to net7 (one needs to check).

> although I am fine compiling against net7 exclusively anyways as any .net maui should upgrade regardless in my opinion.

Either that, or we could do multi-targeting with net6 and net7, as proposed by @twenzel. Both would be ok with me. However, I have little hope that more multi-targeting with help with our iOS bug, since @RobertLinak's investigation seems to indicate that the problem is actually introduced by the multi-targeting with Xamarin and net6.

> as for the ios issue... i havent' looked into it yet, but should be able to this week or next week sometime... i mean should just be able to pull up the nuget and see. I downloaded it and it was all looking good in the nuget package explorer

If have created PR #638 to update the .NET SDK in GHA to v7 (but still targeting only net6 for now). The build went through fine, at least, and there is a nupkg artifact to try out. @RobertLinak, could you check if this makes any difference wrt the iOS bug?

RobertLinak:
> If have created PR #638 to update the .NET SDK in GHA to v7 (but still targeting only net6 for now). The build went through fine, at least, and there is a nupkg artifact to try out. @RobertLinak, could you check if this makes any difference wrt the iOS bug?

Yes #638 fixes the iOS issue for me

janusw:
> > If have created PR #638 to update the .NET SDK in GHA to v7 (but still targeting only net6 for now). The build went through fine, at least, and there is a nupkg artifact to try out. @RobertLinak, could you check if this makes any difference wrt the iOS bug?
> 
> Yes #638 fixes the iOS issue for me

It does? Awesome 😆 🎉

PR #638, opened 2022-11-15
GitHub Actions: update to .NET 7

janusw:
* and update the version on some actions
* the final .NET version 7 has been released recently


janusw:
Apparently this fixes the iOS problems with MAUI (as reported by @RobertLinak in #614), so I'll merge it.

Issue #356, opened 2019-04-24
BLE in Windows Application (UWP)


jvhgamer:
Following the readme, under header "## Usage" we are stuck at initializing the plug-in.

So-
How to address portable version of this assembly in a Windows application?
or/and
Where to initialize Mvx.Resolve in a Windows Application?

## Steps to reproduce

1. Add Plugin.BLE / MvvmCross.Plugin.BLE via NuGet to Windows Application VS solution.

2. Attempt to initialize with either the 'Vanilla' or 'MvvmCross' calls to init BLE.

3. Error(s) will produce

## Expected behavior
Tell us what should happen
Run

## Actual behavior
Tell us what happens instead
Errors out

### Crashlog

If something causes an exception paste full stack trace + Exception here

```
//**Vanilla**
try
{
    var ble = CrossBluetoothLE.Current;
    var adapter = CrossBluetoothLE.Current.Adapter;
}
catch (Exception ex)
{
    Console.WriteLine(ex.Message);
}
```

  | Name | Value | Type
-- | -- | -- | --
  | Message | "This functionality is not implemented in the portable version of this assembly.  You should reference the NuGet package from your main application project in order to reference the platform-specific implementation." | string

  | Name | Value | Type
-- | -- | -- | --
  | StackTrace | "   at Plugin.BLE.CrossBluetoothLE.get_Current()\r\n   at MyWindowsApp.Source.Logic.InitBLE() in C:\\...\\MyWindowsApp\\Source\\Logic.cs:line 116" | string

```
//**MvvmCross**
try
{
    var ble = Mvx.Resolve<IBluetoothLE>();
    var adapter = Mvx.Resolve<IAdapter>();
}
catch (Exception ex)
{
    Console.WriteLine(ex.Message);
}
```

  | Name | Value | Type
-- | -- | -- | --
  | Message | "Object reference not set to an instance of an object." | string

  | Name | Value | Type
-- | -- | -- | --
  | StackTrace | "   at MvvmCross.Platform.Mvx.Resolve[TService]()\r\n   at MyWindowsApp.Source.Logic.InitBLE() in C:\\...\\MyWindowsApp\\Source\\Logic.cs:line 127" | string

## Configuration

**Version of the Plugin:** e.g. 1.0.0 / 1.0.1-alpha3
1.3.0

[...]

PR #640, opened 2022-11-23
Fix problems with MAUI on Windows

janusw:
This PR does a couple of things:
* It updates Microsoft.Toolkit.Uwp.Connectivity from 6.1 to 7.1 (for the Xamarin.UWP case, for MAUI on Windows we already have CommunityToolkit.WinUI.Connectivity v7.1).
* It updates the minimum target platform version for Xamarin.UWP to 10.0.17763 (which is necessary for the above Toolkit version and matches what we already have for MAUI).
* It fixes the NullReferenceException from #356 by obtaining the DispatcherQueue on the UI thread.
* It also updates BLE.Client.UWP a bit.

I verified with BLE.Client.UWP that scanning on Windows works well, and I strongly assume that things should also work with MAUI now (this still needs confirmation, though).

### DIFF
--- a/Source/BLE.Client/BLE.Client.UWP/BLE.Client.UWP.csproj
+++ b/Source/BLE.Client/BLE.Client.UWP/BLE.Client.UWP.csproj
@@ -12,7 +12,7 @@
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
     <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.19041.0</TargetPlatformVersion>
-    <TargetPlatformMinVersion>10.0.16299.0</TargetPlatformMinVersion>
+    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
     <MinimumVisualStudioVersion>14</MinimumVisualStudioVersion>
     <FileAlignment>512</FileAlignment>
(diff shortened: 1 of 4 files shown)


janusw:
> I verified with BLE.Client.UWP that scanning on Windows works well, and I strongly assume that things should also work with MAUI now (this still needs confirmation, though).

By now I have checked via [MauiXAMLBluetoothLE](https://github.com/TimoKinnunen/MauiXAMLBluetoothLE) that scanning works also with MAUI on Windows, even when doing unfiltered scans:
```
diff --git a/MauiXAMLBluetoothLE/Services/BluetoothLEService.cs b/MauiXAMLBluetoothLE/Services/BluetoothLEService.cs
index 65131ba..dc0f36a 100644
--- a/MauiXAMLBluetoothLE/Services/BluetoothLEService.cs
+++ b/MauiXAMLBluetoothLE/Services/BluetoothLEService.cs
@@ -42,7 +42,7 @@ public class BluetoothLEService
                     await ShowToastAsync($"Found {systemDevice.State.ToString().ToLower()} device {systemDevice.Name}.");
                 }
             }
-            await Adapter.StartScanningForDevicesAsync(HeartRateUuids.HeartRateServiceUuids);
+            await Adapter.StartScanningForDevicesAsync();
         }
         catch (Exception ex)
         {
```

janusw:
Since everything seems to work well, I have merged this into master now, and will prepare a beta.2 release with this fix soon.

nefen:
Currently this toolit does not support .Net 7.0.
So, in Maui on net 7 this cannot e installed and Plugin.BLE is unusable for Windows Machine.
Are there any plans to issue an upgrade to support .net 7.0?

Issue #642, opened 2022-11-27
[MAUI - WINDOWS ] BluetoothLE attributes IsOn and IsAvailable return always true independant of the Bluetooth state on the device.


Youldboukhitine:
The BluetoothLE attributes IsOn and isAvailable are not matching with the device state (computer bluetooth peripheral).
These attributes are always to true. 

Is there any other way to check if the Bluetooth peripheral is enabled on the device side ?

MAUI with .NET 7 on Windows 10.x

## Steps to reproduce

1. On windows -- > Settings --> Bluetooth and other devices

2. Turn OFF Bluetooth

3. create a simple app and read the status of a BluetoothLE  variable 

## Expected behavior
The attributes IsOn and IsAvailable shall match with the Bluetooth peripheral state

## Actual behavior
got always true

### Crashlog

N/A

## Configuration

**Version of the Plugin:** 3.0.0-beta-2

**Platform:** Windows 10 OS build 19044.2130

**Device:** : DELL Laptop

smsissuechecker:
Hi @Youldboukhitine,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

janusw:
See `BleImplementation.GetInitialStateNative`: https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/f474528254bf227dc984477a9367aa088d17c301/Source/Plugin.BLE/Windows/BleImplementation.cs#L29-L40

I guess one should at least check [BluetoothLEHelper.IsBluetoothLESupported](https://learn.microsoft.com/en-us/dotnet/api/microsoft.toolkit.uwp.connectivity.bluetoothlehelper.isbluetoothlesupported?view=win-comm-toolkit-dotnet-6.1#microsoft-toolkit-uwp-connectivity-bluetoothlehelper-isbluetoothlesupported) in addition.

Youldboukhitine:
@janusw Thanks for the info.

janusw:
Why did you close this issue? It's a valid bug report, and I simply wanted to point out the relevant piece of code. Despite the comment there, I think the behavior can and should be improved ...

Youldboukhitine:
I don't to waste the repo with something that you already considered. 
That's ok for me to leave it open :)

DamianSuess:
[...]

PR #652, opened 2023-01-29
UWP IsAvailable and Dynamic Status

DamianSuess:
As a user, I would like to have UWP automatically inform me of the Radio's state change (`IsAvailable`) as well as properly informed of its availability on initialization of the library. 

This PR provides said requests as per Issue #642.

## Feature Includes:

* UWP/WinUI implemented Radio StateChanged.
* Added EditorConfig for maintaining spaces/indentation

p.s.
I hope you don't mind I included the EditorConfig file. If it's an issue or out of scope, I can remove it. The intent was to help keep consistency with developer environments and this open-source project.  I removed the style ruleset which I usually keep in there.. food for thought, it might be good to add some ;)


DamianSuess:
On line 41 of [BleImplementation.cs](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/pull/652/files#diff-d2f884ba256ca63261a0e415c4f1c1062d4b40a472c905781313a42cafa0642e), I placed in there a 100ms wait. This is because the async Task falls through too quickly and the `_radio` object isn't created yet.

If anyone has a cleaner suggestion for this workaround, please, please let me know.

DamianSuess:
With regard to the `Task.Run(..).Wait(20)`, it was found that during the app's startup `BleImplementation.State` gets the async operations for Radio can get a few clock cycles too late. Thus the `GetInitialStateNative()` returns "_Unknown_" and the app's ViewModel's is still constructing its event listener for `StateChanged` and doesn't get receive the updated status.

I tried using 10ms and lower, but it was flaky during debugging.

DamianSuess:
Sorry for the constant re-pushes. I was trying to squeeze out every clock cycle. Feel free to squish this merge if accepted 😉

DamianSuess:
Any word if this needs refactored or is good-to-go?

janusw:
Hi Damian, thanks a lot for your contribution, it is highly appreciated. I'll try to review it within the next week ...

Source/Plugin.BLE/Windows/BleImplementation.cs
```diff
@@ -21,30 +23,67 @@ public class BleImplementation : BleImplementationBase
...
 
         protected override void InitializeNative()
         {
             //create local helper using the app context
             var localHelper = BluetoothLEHelper.Context;
             _bluetoothHelper = localHelper;
+
+            // Wait for the async radio response to catch up, otherwise it will be null
+            // If it falls through, the StateChanged event doesn't get fired in time
+            Task.Run(InitRadioStateAsync).Wait(10);
```
janusw, line 53: I don't quite understand your argumentation regarding the `Wait` here. You changed the timespan from 100ms to 20ms to 10ms, but it's completely unclear how these values are motivated, or why one is better than the other. Wouldn't it be best to wait indefinitely (without a timeout parameter) until the state initialization has fully completed? Or am I missing something?

janusw (commented):
So, finally some reviewing (sorry that it took me so long) ...

Source/Plugin.BLE/Windows/BleImplementation.cs
DamianSuess: Though I admit to not being a fan of hard timeouts in code, the provided implementation was to get things rolling while delivering a fix.  In the case of 100ms -> 10ms was merely squeezing out the most minimum clock cycles.

Normally in my own systems, I would set an uninitialized state and wait for the event to come back (i.e. `Radio_StateChanged(..)`) to set it accordingly. However, with UWP system having or not having a BLE radio, I'm sure of the effects of doing something like `WaitFof(() => State != BluetoothState.Unknown, MaxTimeoutHere);`

Although, perhaps, this is exactly what we could use. What are your thoughts?

janusw: Well, my first suggestion would be to change line 53 to ...
```
Task.Run(InitRadioStateAsync).Wait();
```
... in order to make sure that we're waiting for the task to fully run through. If you pass any `timeout` parameter, you always risk that the execution continues before the task has finished and the initialization is not done yet, so that we don't have a well-defined initial state.

Or is there any reason why this would not be a good idea?

DamianSuess: Made the change & pushed it up. So far, so good. It's been a while since this PR was opened, and I can't reproduce the reason for needing the 10ms.

Do you mind giving this branch a shot on your machine w/ toggling the BLE radio on and off. Just want to make sure that things are good to go here.

janusw: Sorry, I haven't managed to verify the functionality yet (Windows is really not my native habitat 😉, plus: the sample client currently does not seem to show the radio state).

Is there any easy way to check the changed behavior on your branch? If not, maybe one could extend the sample client to display the radio state in some way?

Have you tested it yourself by now? (Or maybe someone else has?)

AskBojesen:
@DamianSuess : Why don't you use the methods in the `windows.devices.enumeration` API ?

https://learn.microsoft.com/en-us/uwp/api/windows.devices.enumeration

janusw:
This PR is outdated by now and has conflicts with the current master branch. The problem may even by fixed via #746. If not, it's probably the best to open a new PR. I'll close this one for now.

Issue #670, opened 2023-04-09
[Issue] If remove device disconnects during communication, event is not passed to application


chrimaka:
If during communication the remote device disconnects, this is not passed correctly to the application back

## Steps to reproduce

1. Establish communication with remote BLE Device. 
2. Send a read or write command which causes the BLE device to disconnected. 
3. Application still "thinks" it is connect to the remote BLE Device. 

## Expected behavior
The Disconnect Event should have been passt the application. 

## Actual behavior
Read or Write Command return, but no disconnect event. 

### Crashlog

If something causes an exception paste full stack trace + Exception here

## Configuration

**Version of the Plugin:** 2.1.3

**Platform:** Android in General

smsissuechecker:
Hi @chrimaka,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

janusw:
This should be fixed by #671. I'm releasing 3.0.0-beta.3 today (which includes this fix).

PR #671, opened 2023-04-09
Adding propagating disconnect Event if happening during communcation …

chrimaka:
… (#670)

By removing the 'break' the "inform pending task" is happening for sure, that the applications become award of the disconnect.


janusw (approved):
LGTM. Thanks for the contribution!

Issue #675, opened 2023-04-18
[Fearture Request] Querying the adapter's BLE5 capabilities


janusw:
It would be great if there was a way to query the BLE adapter's capabilites, in particular regarding Bluetooth 5 support.

I'm primarily interested in querying support for the following two features (but there are certainly more).

### Extended Advertisements

Android: https://developer.android.com/reference/android/bluetooth/BluetoothAdapter#isLeExtendedAdvertisingSupported()
iOS: https://developer.apple.com/documentation/corebluetooth/cbcentralmanagerfeature/cbcentralmanagerfeatureextendedscanandconnect

### Coded PHY Support

Android: https://developer.android.com/reference/android/bluetooth/BluetoothAdapter#isLeCodedPhySupported()
iOS: not available so far? (so we'd always report `false`?)

### Further features

(from the Android API, not sure which of those are available on iOS as well)

* 2M PHY support
* LE audio support
* periodic advertising
* multi advertisement
* offloaded filtering
* offloaded scan batching
* ... ?

Supporting such feature queries in Plugin.BLE only really makes sense if they are available on more than one platform (but if someone really needs them on Android only, they certainly can be added as well).

smsissuechecker:
Hi @janusw,

I'm the friendly issue checker.
It seems like (100.00 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

If you think it's fine to make an exception, just ignore this message.
**But if you think it was a mistake to delete the template, please close the issue and create a new one.**

Thanks!

PR #679, opened 2023-04-25
Querying adapter capabilities: extended advertisements & coded phy

janusw:
This PR implements the major parts of issue #675, adding two new methods to the IAdapter interface:

* `supportsExtendedAdvertising`
* `supportsCodedPHY`

The first one has implementations on Android and iOS (and returns false on other platforms, such as MacOS and Windows), the second one is implemented only on Android.


janusw (commented):
I tested this on Android and iOS, and it seems to work as expected. Also the CI build runs through fine (without any additional warnings), so I'm going to merge this one.

Issue #686, opened 2023-05-06
Provide IsConnectable in advertised/discovered device 


midoragh:
To help us fix your issue, please provide the information in the below template. If something causes a crash, provide as much information as you can gather.
Just imagine: we do not know what you are doing! 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.

## Steps to reproduce

None

## Expected behavior
The device should provide IsConnectable flag. This bool is available on Android in ScanResult and on iOS in CBAdvertisement.

## Actual behavior
The flag is not available

### Crashlog

no crash

## Configuration

**Version of the Plugin:** 1.2.3

**Platform:** Android 10, Api 29

**Device:** Wiko

smsissuechecker:
Hi @midoragh,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

midoragh:
I would prefer this device solution instead of storing the value in AdvertisementRecordType.IsConnectable as done in iOS.

janusw:
> The device should provide IsConnectable flag. This bool is available on Android in ScanResult and on iOS in CBAdvertisement.

Sounds like a reasonable enhancement.



> I would prefer this device solution instead of storing the value in AdvertisementRecordType.IsConnectable as done in iOS.

Makes sense, and should be possible. Related comment:

https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/2e8615de6515fa2593ee39004a84116ec636b8e3/Source/Plugin.BLE/Shared/AdvertisementRecord.cs#L178-L181


PRs welcome ;)

midoragh:
I will create PR. I already tested that Androi provides a good value by adding it 0xAA to the advertised records.

midoragh:
Seems I do not have write access to create a branch.

midoragh:
VS2022 Windows Community
Error; The target "GetApkRecipeFile" does not exist in the project.; BLE.Client.Droid; QWERTY\dble-IsConnectable\dble-IsConnectable\Source\MvvmCross.Plugins.BLE\MvvmCross.Plugins.BLE.csproj

janusw:
[...]

PR #688, opened 2023-05-10
IsConnectable

midoragh:
Added IsConnectable and CanIsConnectable to IDevice. Removed old handling of IsConnectable in the apple case. Only Android versions < Lollipop do not support IsConnectable.

### DIFF
--- a/Source/BLE.Client/BLE.Client.Droid/BLE.Client.Droid.csproj
+++ b/Source/BLE.Client/BLE.Client.Droid/BLE.Client.Droid.csproj
@@ -96,6 +96,8 @@
     <ProjectReference Include="..\..\MvvmCross.Plugins.BLE\MvvmCross.Plugins.BLE.csproj">
       <Project>{83f1dffb-a653-45dd-be81-c02374c6db07}</Project>
       <Name>MvvmCross.Plugins.BLE</Name>
+      <TargetPlatformIdentifier>Android</TargetPlatformIdentifier>
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
     </ProjectReference>
     <ProjectReference Include="..\..\Plugin.BLE\Plugin.BLE.csproj">
       <Project>{951ed11d-d026-449c-90dc-5027bc41fb3b}</Project>
--- a/Source/Plugin.BLE/Android/Adapter.cs
+++ b/Source/Plugin.BLE/Android/Adapter.cs
@@ -247,7 +247,7 @@ namespace Plugin.BLE.Android
             var macBytes = deviceGuid.ToByteArray().Skip(10).Take(6).ToArray();
             var nativeDevice = _bluetoothAdapter.GetRemoteDevice(macBytes);
 
-            var device = new Device(this, nativeDevice, null, 0, new byte[] { });
+            var device = new Device(this, nativeDevice, null);
 
             await ConnectToDeviceAsync(device, connectParameters, cancellationToken);
             return device;
@@ -265,7 +265,7 @@ namespace Plugin.BLE.Android
 
             var bondedDevices = _bluetoothAdapter.BondedDevices.Where(d => d.Type == BluetoothDeviceType.Le || d.Type == BluetoothDeviceType.Dual);
 
-            return connectedDevices.Union(bondedDevices, new DeviceComparer()).Select(d => new Device(this, d, null, 0)).Cast<IDevice>().ToList();
+            return connectedDevices.Union(bondedDevices, new DeviceComparer()).Select(d => new Device(this, d, null)).Cast<IDevice>().ToList();
         }
 
         public override IReadOnlyList<IDevice> GetKnownDevicesByIds(Guid[] ids)
@@ -334,10 +334,9 @@ namespace Plugin.BLE.Android
             {
                 Trace.Message("Adapter.LeScanCallback: " + bleDevice.Name);
 
-                _adapter.HandleDiscoveredDevice(new Device(_adapter, bleDevice, null, rssi, scanRecord));
+                _adapter.HandleDiscoveredDevice(new Device(_adapter, bleDevice, null, rssi, scanRecord)); // No IsConnectable!
             }
         }
-
 
         public class Api21BleScanCallback : ScanCallback
         {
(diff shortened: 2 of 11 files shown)


Source/BLE.Client/BLE.Client.Droid/BLE.Client.Droid.csproj
```diff
@@ -96,6 +96,8 @@
     <ProjectReference Include="..\..\MvvmCross.Plugins.BLE\MvvmCross.Plugins.BLE.csproj">
       <Project>{83f1dffb-a653-45dd-be81-c02374c6db07}</Project>
       <Name>MvvmCross.Plugins.BLE</Name>
+      <TargetPlatformIdentifier>Android</TargetPlatformIdentifier>
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
```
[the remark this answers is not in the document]
midoragh: Nothing, I'm pretty sure thar VS decided to do this modification.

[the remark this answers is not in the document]
janusw: Ah, ok. In any case, it needs to be removed again, in order to fix the build.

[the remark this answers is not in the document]
janusw: Fixed by 5f2c327ede75534913010b7695faae0bc9829e7b. Seems to build now. Thanks! :+1:

janusw:
One should mention that this is supposed to fix #686.

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
```diff
@@ -106,5 +106,18 @@ public interface IDevice : IDisposable
...
+        /// Gets the information if the device has hinted during advertising that the device is connectable.
+        /// This information is not pat of an advertising record. It's determined from the PDU header.
+        /// Check CanIsConnectable to verify that the device supports IsConnectable.
+        /// </summary>
+        bool IsConnectable { get; }
+
+        /// <summary>
+        /// True, if device supports IsConnectable else False
+        /// </summary>
+        bool CanIsConnectable { get; }
```
janusw, line 121: One rather 'cosmetic' comment here: `CanIsConnectable` does not seem to be a very good name to me (sounds a bit weird ;). How about calling it `SupportsIsConnectable` instead?

Source/Plugin.BLE/Android/Adapter.cs
```diff
@@ -384,7 +383,7 @@ public override void OnScanResult(ScanCallbackType callbackType, ScanResult resu
                     records.Add(new AdvertisementRecord(AdvertisementRecordType.ServiceData, result.ScanRecord.ServiceData));
                 }*/
 
-                var device = new Device(_adapter, result.Device, null, result.Rssi, result.ScanRecord.GetBytes());
+                var device = new Device(_adapter, result.Device, null, result.Rssi, result.ScanRecord.GetBytes(), result.IsConnectable);
```
janusw, line 386: There is a warning here:
```
This call site is reachable on: 'Android' 21.0 and later. 'ScanResult.IsConnectable' is only supported on: 'android' 26.0 and later.
```

Source/Plugin.BLE/Android/Device.cs
```diff
@@ -412,5 +414,10 @@ protected override bool UpdateConnectionIntervalNative(ConnectionInterval interv
                 throw new Exception($"Update Connection Interval fails with error. {ex.Message}");
             }
         }
+
+        public override bool IsConnectable { get; protected set; }
+
+        public override bool CanIsConnectable { get => (Build.VERSION.SdkInt >= BuildVersionCodes.Lollipop); }
```
janusw, line 420: Why do you use `Lollipop` here? That's Android 5 / API level 21. It actually seems like `IsConnectable` is only supported on API level 26 and above (Android 8 / Oreo). See https://developer.android.com/reference/android/bluetooth/le/ScanResult#isConnectable()

janusw (commented):
Thanks for this contribution! Overall it all sounds very reasonable and seems like a good addition to the API. I only have a couple of minor comments ...

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
midoragh: Choosed Can... because of CanRead, CanWrite, CanUpdate of characteristics.
But Supports... is ok, I will change it.

Source/Plugin.BLE/Android/Device.cs
midoragh: Same reason as for the previous issue. I just looked when the interface provided ScanResult.

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
janusw: Well, yeah, for the other cases 'can' makes more sense grammatically, for 'IsConnectable' it really doesn't.
Thanks for your consideration. :)

Source/Plugin.BLE/Android/Device.cs
janusw: I don't understand. Which interface do you mean? Can you elaborate?
As you can see from the link above, the Android API seems to support the `isConnectable` method since API level 26, no?

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
```diff
@@ -106,5 +106,18 @@ public interface IDevice : IDisposable
         /// <returns>True if the update request was sucessfull. On iOS it will always return false.</returns>
         /// <param name="interval">The requested interval (High/Low/Normal)</param>
         bool UpdateConnectionInterval(ConnectionInterval interval);
+
+
+        /// <summary>
+        /// Gets the information if the device has hinted during advertising that the device is connectable.
+        /// This information is not pat of an advertising record. It's determined from the PDU header.
+        /// Check CanIsConnectable to verify that the device supports IsConnectable.
```
janusw, line 114: There is another reference to 'CanIsConnectable' here that should be changed to 'SupportsIsConnectable'.

Source/Plugin.BLE/Android/Adapter.cs
janusw: The warnings seem to be gone by now. Thanks for fixing!

Issue #713, opened 2023-07-14
Index was outside the bounds of the array


vitalii-vov:
The app crashes when trying to scan devices.
This happens on the client device. Repeats only for one user. I managed to get logs from him by sending him a debug version of the program.

## Steps to reproduce

1. Start scan devices
2. Wait
3. App crushed after some seconds

## Expected behavior
App should work

## Actual behavior
App crashes

### Crashlog
```
Index was outside the bounds of the array.
  at Plugin.BLE.Android.Device.ParseScanRecord (System.Byte[] scanRecord) [0x00020] in <e84f194db96045279f39298a58a796ea>:0 
  at Plugin.BLE.Android.Device..ctor (Plugin.BLE.Android.Adapter adapter, Android.Bluetooth.BluetoothDevice nativeDevice, Android.Bluetooth.BluetoothGatt gatt, System.Int32 rssi, System.Byte[] advertisementData) [0x00017] in <e84f194db96045279f39298a58a796ea>:0 
  at Plugin.BLE.Android.Adapter+Api21BleScanCallback.OnScanResult (Android.Bluetooth.LE.ScanCallbackType callbackType, Android.Bluetooth.LE.ScanResult result) [0x00026] in <e84f194db96045279f39298a58a796ea>:0 
  at Android.Bluetooth.LE.ScanCallback.n_OnScanResult_ILandroid_bluetooth_le_ScanResult_ (System.IntPtr jnienv, System.IntPtr native__this, System.Int32 native_callbackType, System.IntPtr native_result) [0x00011] in <5437bca4f0e742d0a18eb9a3847ccdb6>:0 
  at (wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.1(intptr,intptr,int,intptr)
```

## Configuration

**Version of the Plugin:** 2.1.3

**Platform:** Android 13 (One UI 5.1)

**Device:** Samsung Galaxy S21 FE 5G, model: SM-G990E/DS

smsissuechecker:
Hi @vitalii-vov,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

vitalii-vov:
The cause of the fall of the application was a wrist bluetooth watch. After the clock was turned off, the application stopped crashing. After turning on the clock, the application no longer crashes.
It turns out that an incorrectly working device just nearby can break the application.
[...]

Issue #567, opened 2022-03-11
Scan causes crash on Android detected by AppCenter


sandros06:
Hi, 

I post here because I received on AppCenter of my application a crash error from Plugin.BLE, I don't know how and why this error is triggered but I wish with the Xamarin exception stack thah you can help me

## Steps to reproduce

Nothing sorry

### Crashlog

```
Xamarin Exception Stack:
System.ArgumentException: length
  at System.Array.Copy (System.Array sourceArray, System.Int32 sourceIndex, System.Array destinationArray, System.Int32 destinationIndex, System.Int32 length) [0x000b4] in <5a88e350e0674dd7b8ddba06471cdfe2>:0
  at Plugin.BLE.Android.Device.ParseScanRecord (System.Byte[] scanRecord) [0x00066] in <17887b5227df4ea9ab9b317d496489c6>:0
  at Plugin.BLE.Android.Device..ctor (Plugin.BLE.Android.Adapter adapter, Android.Bluetooth.BluetoothDevice nativeDevice, Android.Bluetooth.BluetoothGatt gatt, System.Int32 rssi, System.Byte[] advertisementData) [0x00017] in <17887b5227df4ea9ab9b317d496489c6>:0
  at Plugin.BLE.Android.Adapter+Api21BleScanCallback.OnScanResult (Android.Bluetooth.LE.ScanCallbackType callbackType, Android.Bluetooth.LE.ScanResult result) [0x00026] in <17887b5227df4ea9ab9b317d496489c6>:0
  at Android.Bluetooth.LE.ScanCallback.n_OnScanResult_ILandroid_bluetooth_le_ScanResult_ (System.IntPtr jnienv, System.IntPtr native__this, System.Int32 native_callbackType, System.IntPtr native_result) [0x00011] in <950ad56dbf9942d1a0c82aaea5a2cf2a>:0
  at (wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.49(intptr,intptr,int,intptr)

```
## Configuration

**Version of the Plugin:** 2.1.1
**Android:** 9
**Android Build:** 50.2.A.3.77
**Manufacturer:** Sony
**Model:** Xperia XA2 (H3113)

smsissuechecker:
Hi @sandros06,

I'm the friendly issue checker.
It seems like (50.00 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

[...]

PR #719, opened 2023-07-21
Fixed crash with incorrect data in advertisement data on android

vitalii-vov:
fix for https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/issues/713
and https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/issues/567

### DIFF
--- a/Source/Plugin.BLE/Android/Device.cs
+++ b/Source/Plugin.BLE/Android/Device.cs
@@ -46,7 +46,7 @@ namespace Plugin.BLE.Android
         {
             Update(nativeDevice, gatt);
             Rssi = rssi;
-            AdvertisementRecords = ParseScanRecord(advertisementData);
+            AdvertisementRecords = ParseScanRecord(advertisementData, nativeDevice);
             IsConnectable = isConnectable;
             _gattCallback = new GattCallback(adapter, this);
         }
@@ -256,7 +256,7 @@ namespace Plugin.BLE.Android
             return new Guid(deviceGuid);
         }
 
-        public static List<AdvertisementRecord> ParseScanRecord(byte[] scanRecord)
+        public static List<AdvertisementRecord> ParseScanRecord(byte[] scanRecord, BluetoothDevice nativeDevice)
         {
             var records = new List<AdvertisementRecord>();
 
@@ -319,7 +319,7 @@ namespace Plugin.BLE.Android
             }
             catch(Exception)
             {
-                Trace.Message($"Filed to parse advertisement data: {scanRecord.ToHexString()}");
+                Trace.Message("Failed to parse advertisementData. Device address: {0}, Data: {1}", nativeDevice.Address, scanRecord.ToHexString());
                 //There may be a situation where scanRecord contains incorrect data.
                 //We return an empty list, even if some of the data is recognized,
                 //because if the scanRecord is invalid, then we cannot be sure of the data.
--- a/Source/Plugin.BLE/Android/Device.cs
+++ b/Source/Plugin.BLE/Android/Device.cs
@@ -46,7 +46,7 @@ namespace Plugin.BLE.Android
         {
             Update(nativeDevice, gatt);
             Rssi = rssi;
-            AdvertisementRecords = ParseScanRecord(advertisementData, nativeDevice);
+            AdvertisementRecords = ParseScanRecord(advertisementData);
             IsConnectable = isConnectable;
             _gattCallback = new GattCallback(adapter, this);
         }
@@ -256,7 +256,7 @@ namespace Plugin.BLE.Android
             return new Guid(deviceGuid);
         }
 
-        public static List<AdvertisementRecord> ParseScanRecord(byte[] scanRecord, BluetoothDevice nativeDevice)
+        public List<AdvertisementRecord> ParseScanRecord(byte[] scanRecord)
         {
             var records = new List<AdvertisementRecord>();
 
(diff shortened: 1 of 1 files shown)


Source/Plugin.BLE/Android/Device.cs
```diff
@@ -262,58 +263,68 @@ public static List<AdvertisementRecord> ParseScanRecord(byte[] scanRecord)
...
+                    records.Add(record);
+
+                    //Advance
+                    index += length;
+                }
+                return records;
+            }
+            catch(Exception)
+            {
+                Trace.Message($"Filed to parse advertisement data: {scanRecord.ToHexString()}");
```
janusw, line 322: Typo here: "Filed" -> "Failed"

Source/Plugin.BLE/Android/Device.cs
```diff
@@ -262,58 +263,68 @@ public static List<AdvertisementRecord> ParseScanRecord(byte[] scanRecord)
...
+                //There may be a situation where scanRecord contains incorrect data.
+                //We return an empty list, even if some of the data is recognized,
+                //because if the scanRecord is invalid, then we cannot be sure of the data.
+                return new List<AdvertisementRecord>();
```
janusw, line 326: Maybe this should better be `return records`? If an advertisement contains multiple records, and the failure occurs only in a later one, we could at least return the earlier ones, which were ok. Right?

Source/Plugin.BLE/Android/Device.cs
janusw: In addition: Maybe your trace output should also contain the device address, so that it is easier to see which device is causing the issue?

janusw (changes requested):
Overall this is very reasonable, I think. I only have some minor comments (see below) ...

Source/Plugin.BLE/Android/Device.cs
vitalii-vov: Yes you are right

Source/Plugin.BLE/Android/Device.cs
vitalii-vov: I did so at the beginning, but then I thought that we could not be sure that the package was not damaged at the beginning or in the middle.
That is, if we extract the middle from the data "02-01-1A-02-0A-08-0C-FF-4C-00-10-07-35-1F-69-21-B3-32-08", for example, starting from index 5 (02-01-1A-02-0A-4C-00-10-07-35-1F-69-21-B3-32- 08), then we will get incorrect data already inside the recognized records.
If the package is in the wrong format, then in theory it can be broken anywhere.
However, I can rewrite the code to return post values.

vitalii-vov: In the example I just gave, there will not be an exception, since the loop will be interrupted at the line
~~~
if (!Enum.IsDefined(typeof(AdvertisementRecordType), type))
{
    Debug.WriteLine("Advertisment record type not defined: {0}", type);
    break;
}
~~~
which is also probably not correct, since it will contain bad data?
Perhaps in this place, too, you need to return an empty array?

Source/Plugin.BLE/Android/Device.cs
vitalii-vov: How would you like to pass the device address to the function?
For example like this:
~~~C#
public static List<AdvertisementRecord> ParseScanRecord(byte[] scanRecord, BluetoothDevice nativeDevice)
{
   // ...
   Trace.Message("Failed to parse advertisementData: {0} device address: {1}", scanRecord.ToHexString(), nativeDevice.Address);
}
~~~

janusw: > How would you like to pass the device address to the function?

This is not necessary. The `Device` class has a member `NativeDevice`, so you can directly access `NativeDevice.Address` in the method `ParseScanRecord`.

Source/Plugin.BLE/Android/Device.cs
janusw: > I did so at the beginning, but then I thought that we could not be sure that the package was not damaged at the beginning or in the middle.

Well, we never have any guarantee that the data is "correct". We just keep parsing as long as it's plausible.


> However, I can rewrite the code to return recognized messages.

Yeah, I think that's the best we can do. One could simply pull the `return records` statement out of the `try` block and put it after the `catch` block (which should then not contain a `return` statement itself).

janusw: > In the example I just gave, there will not be an exception, since the loop will be interrupted at the line
> 
> ```
> if (!Enum.IsDefined(typeof(AdvertisementRecordType), type))
> {
>     Debug.WriteLine("Advertisment record type not defined: {0}", type);
>     break;
> }
> ```
> 
> which is also probably not correct, since it will contain bad data. Perhaps in this place, too, you need to return an empty array?

No, IMHO we should parse as much as possible, and stop only once there is an obvious problem. I don't think it's a very likely scenario that the beginning of a record is missing. I assume that most 'bad' advertisements would be sent out by 'buggy' devices that simply transmit non-standard advertisements (intentionally or not).

Source/Plugin.BLE/Android/Device.cs
```diff
@@ -255,65 +256,75 @@ private Guid ParseDeviceId()
             return new Guid(deviceGuid);
         }
 
-        public static List<AdvertisementRecord> ParseScanRecord(byte[] scanRecord)
+        public static List<AdvertisementRecord> ParseScanRecord(byte[] scanRecord, BluetoothDevice nativeDevice)
```
janusw, line 259: As mentioned above, the `nativeDevice` parameter is unnecessary, since the `Device` class has a `NativeDevice` member.

Do you want to fix the remaining comments yourself, or would prefer that I take over?

vitalii-vov: I thought that passing the device through a parameter is a cleaner option, since it does not depend on the order of the function calls.
I hope I did everything right. But you can continue as you see fit.

Source/Plugin.BLE/Android/Device.cs
vitalii-vov: I think you have more experience with this. On this you are right

Source/Plugin.BLE/Android/Device.cs
janusw: I actually missed that `ParseScanRecord` is declared as a static method, but simply removing the `static` (as you did) is certainly a good solution. Looks pretty good by now.

janusw (approved):
Looks good to me now. I will merge by tomorrow, if you happy with the current state as well, @vitalii-vov.

janusw:
Merged to master. Thanks for the contribution, @vitalii-vov!

vitalii-vov:
Hello @janusw 
Can you please tell me when do you plan to release the nuget package?

Issue #620, opened 2022-09-05
Impossible to reconnect device in UWP


andrey-val-rodin:
If you connect device and then disconnect it, trying to connect the device again will result in unhandled exception.

## Steps to reproduce

1. Run sample app
2. Connect some device->device connected->Ok
3. Click back arrow in top left corner
4. Click on device again and press Disconnect
5. Try to connect the device ->UnhandledException event fired

## Expected behavior
Second connection is successful

## Actual behavior
Unhandled exception

### Crashlog

-		Exception	{"The object has been closed. (Exception from HRESULT: 0x80000013)"}	System.Exception {System.ObjectDisposedException}
+		Data	{System.Collections.ListDictionaryInternal}	System.Collections.IDictionary {System.Collections.ListDictionaryInternal}
		HResult	-2147483629	int
		HelpLink	null	string
+		InnerException	null	System.Exception
		Message	"The object has been closed. (Exception from HRESULT: 0x80000013)"	string
		ObjectName	""	string
		Source	null	string
		StackTrace	null	string
		TargetSite	null	System.Reflection.MethodBase

## Configuration

**Version of the Plugin:** 2.2.0 Pre5

**Platform:** UWP

**Device:** desktop computer with Windows 11 (build 22000)

smsissuechecker:
Hi @andrey-val-rodin,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

janusw:
Could this be related to the disconnection issues of #423, #528 and #536? Or is it a separate problem? Does the disconnection operation actually succeed in your case?

Does this still occur with 3.0.0-beta.2?

janusw:
Fixed via #721 in [3.0.0-beta.5](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/releases/tag/3.0.0-beta.5).

Issue #536, opened 2021-09-02
Pre Release 2.2 Disconnect Issue in UWP


sisaacks:
To help us fix your issue, please provide the information in the below template. If something causes a crash, provide as much information as you can gather.
Just imagine: we do not know what you are doing! 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.

## Steps to reproduce

1. Connect to a device

2. Then call _adapter.DisconnectDeviceAsync(connectedDevice);

3.

## Expected behavior
If I call await _adapter.DisconnectDeviceAsync(connectedDevice); I would expect the method to return and the device to be disconnected.  

## Actual behavior
There are two issues, if I call _await _adapter.DisconnectDeviceAsync(connectedDevice);_ the method never returns 
If I remove the await,  then the  device is removed from the _adapter.ConnectedDevices.  The device however is is the _adapater.DiscoveredDevices and its State is still **Connected**.  

### Crashlog

No crashes

## Configuration

**Version of the Plugin:** e.g. Pre Release 2.2.0

Platform is UWP on a windows box

smsissuechecker:
Hi @sisaacks,

I'm the friendly issue checker.
It seems like (25.00 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

If you think it's fine to make an exception, just ignore this message.
**But if you think it was a mistake to delete the template, please close the issue and create a new one.**

Thanks!

janusw:
This is probably a duplicate of #423 and/or #528, right? Does it still occur with 3.0.0-beta.2?

janusw:
Fixed via #721 in [3.0.0-beta.5](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/releases/tag/3.0.0-beta.5).

Issue #528, opened 2021-07-22
Disconnect Doesnt Complete After BT Connecting Successfully On The Second/X Try


JongChern:
To help us fix your issue, please provide the information in the below template. If something causes a crash, provide as much information as you can gather.
Just imagine: we do not know what you are doing! 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.

## Steps to reproduce
0. Using the UWP example provided 
1. Scan
2. Select Scanned Device
3. Power off said Bluetooth Device
4. Connect to said Device while it is powered off
5. Cancel Connect
5. Power on device, and connect
6. Connect is successful
7. Disconnect ~ Disconnect never completes

## Expected behavior
Disconnect should complete

## Actual behavior
Disconnect doesnt complete 

### Crashlog

If something causes an exception paste full stack trace + Exception here

## Configuration

**Version of the Plugin:** e.g. 1.0.0 / 1.0.1-alpha3
https://github.com/xabre/xamarin-bluetooth-le/tree/2.2.0-pre2

**Platform:** e.g. iOS 10.1 / Android 4.4 / ... (including version!!! e.g. Android 5.1 / i0S 10)
Windows 10

**Device:** e.g. HTC Sensation /i Phone 7 ...
Windows PC

smsissuechecker:
Hi @JongChern,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

janusw:
This is possibly a duplicate of #423. Does it still occur with 3.0.0-beta.2?

janusw:
Fixed via #721 in [3.0.0-beta.5](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/releases/tag/3.0.0-beta.5).

Issue #423, opened 2020-04-03
UWP - _adapter.DisconnectDeviceAsync() doesn't work - plugin ver. 2.2.0-pre2


Auto72:
## Steps to reproduce

1. UWP application project with BLE Plugin version: 2.2.0-pre2

2. Open a connection with a device 

await _adapter.ConnectToDeviceAsync(_device);
_myService = await _device.GetServiceAsync(wantedServicetUUID);
_myCharacteristic = await __myService GetCharacteristicAsync(wantedCharacteristicUUID));

_myCharacteristic.ValueUpdated += (o, args) =>
{
    var bytes = args.Characteristic.Value;
};
await _myCharacteristic .StartUpdatesAsync();

3. Close the Connection

await _myCharacteristic.StopUpdatesAsync();
await _adapter.DisconnectDeviceAsync(_device);

## Expected behavior
The connection with the bluetooth LE device is closed.

## Actual behavior
The connection with the bluetooth LE device is NOT closed.
The following command, never ends:

await _adapter.DisconnectDeviceAsync(_device);

### Crashlog

No crash occurs.

## Configuration

Version of the Plugin: 2.2.0-pre2

Platform: Windows 10 1909

Device: Windows 10 1909

smsissuechecker:
Hi @Auto72,

I'm the friendly issue checker.
It seems like (37.50 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

If you think it's fine to make an exception, just ignore this message.
**But if you think it was a mistake to delete the template, please close the issue and create a new one.**

Thanks!

JongChern:
Hi was wondering if there are any immediate plans to get this fix released on nuget

sisaacks:
Hello has this been added to Pre2?  I am still seeing this issue

TStein1984:
@JanNepras @xabre @janusw @sisaacks if have the same issue. Is there some hack to fix it or another solution so solve the problem for UWP? Thanks in advance.

janusw:
Does this still occur with 3.0.0-beta.2?

Auto72:
Still doesn't work with 3.0.0-beta.2.

IDevice device;
ICharacteristic fifoCharacteristic;

The following command now crashes in UWP, even if it is surrounded by a try catch block:

await adapter.DisconnectDeviceAsync(device);

Sometimes also the following command crashes on UWP  (called before the previous one):

[...]

PR #721, opened 2023-07-24
Fix windows connect/disconnect issues #528, #536, #423. #620

TravisRo:
Fixes problems with connecting and disconnecting on Windows:
1) DisconnectDeviceAsync hangs and device remains connected
2) When a connected device loses power, DeviceDisconnected is fired instead of DeviceConnectionLost

Steps to reproduce 1 of 2:
1) Start App
2) Power On Device
3) Start Scanner
4) Select Device
5) Stop Scanner
6) Connect to device
7) Power off device
	* Error: DeviceDisconnected is fired instead of DeviceConnectionLost
8) Power On Device
9) Connect to device
10 Disconnect device
	* Error: Disconnect hangs and device remains connected
	
Steps to reproduce 2 of 2:
1) Start App
2) Power On Device
3) Start Scanner
4) Select Device
5) Stop Scanner
6) Power off device
	* Error: DeviceDisconnected is fired instead of DeviceConnectionLost
7) Connect to device
	* Correct: Device connection fails
8) Power on device
9) Connect to device
10 Disconnect device
	* Error: Disconnect hangs and device remains connected

### DIFF
--- a/Source/Plugin.BLE/Windows/Adapter.cs
+++ b/Source/Plugin.BLE/Windows/Adapter.cs
@@ -19,6 +19,7 @@ using Windows.Devices.Bluetooth.Advertisement;
 using Plugin.BLE.Abstractions;
 using Plugin.BLE.Abstractions.Contracts;
 using Plugin.BLE.Extensions;
+using System.Collections.Concurrent;
 
 namespace Plugin.BLE.UWP
 {
@@ -27,6 +28,12 @@ namespace Plugin.BLE.UWP
         private BluetoothLEHelper _bluetoothHelper;
         private BluetoothLEAdvertisementWatcher _bleWatcher;
         private DispatcherQueue _dq;
+
+        /// <summary>
+        /// Registry used to store device instances for pending operations : disconnect
+        /// Helps to detect connection lost events.
+        /// </summary>
+        private readonly IDictionary<string, IDevice> _deviceOperationRegistry = new ConcurrentDictionary<string, IDevice>();
 
         public Adapter(BluetoothLEHelper bluetoothHelper)
         {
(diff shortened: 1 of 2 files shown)


janusw:
Awesome, fixing three issues in one sweep. 😁
Does it possibly also fix #620?

TravisRo:
The sample app still has problems but I can see why.  When a device is disconnected, DisconnectDeviceNative disposes the IDevice object but this object is cached in ViewModel.  When you try to connect again, it is using a disposed object.  

If you instead use the "Connect to previous", then it works because this uses the ConnectToKnownDeviceAsync function then updates the underlying IDevice object in the ViewModel.

So..  Either you cannot cache IDevice objects (same as services/characteristics) or you can't dispose the IDevice object on disconnect.

My test app assumes the former which is why I don't see this problem with it. I would suggest updating the sample app and your readme.md to clearly state that you cannot reuse any objects after a disconnect or connection lost.

TravisRo:
This second push fixes the #620 issue with the sample program.

janusw:
> This second push fixes the #620 issue with the sample program.

Fantastic! Code-wise it also looks pretty good at first glance.

I'm actually AFK for about ten days now, but I plan to merge this once I'm back. In the meantime, I hope that some of the bug reporters can confirm that the bugs are fixed. @Auto72 @JongChern @sisaacks @andrey-val-rodin @TStein1984 @nathenxbrewer

There are nupkgs built via GitHub Actions that can be used for testing (see the green checkmarks next to the commits).

janusw:
> I'm actually AFK for about ten days now, but I plan to merge this once I'm back. In the meantime, I hope that some of the bug reporters can confirm that the bugs are fixed. @Auto72 @JongChern @sisaacks @andrey-val-rodin @TStein1984 @nathenxbrewer

Huh, unfortunately no feedback here yet, but in any case I'm going to merge this now (as announced).

janusw (approved):

travis012:
I've been beating on it enough to know that it's a step in the right direction.  I still see an issue where disconnect will hang when I set breakpoints in just the right place.  There must be a race condition somewhere but that's a different issue.  I think you can close all the PRs for this one.

PR #724, opened 2023-07-31
Bonding on Android

jefffhaynes:
This enables bonding on Android, which is now required to workaround what appears to be an Android bug causing the pairing dialog to appear twice.  It's possible there should be some sort of IsBondingSupported property as well, as bonding isn't supported on iOS or (I believe) UWP.

### DIFF
--- a/Source/Plugin.BLE/Android/Adapter.cs
+++ b/Source/Plugin.BLE/Android/Adapter.cs
@@ -3,12 +3,15 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Android.App;
 using Android.Bluetooth;
 using Android.Bluetooth.LE;
+using Android.Content;
 using Android.OS;
 using Java.Util;
 using Plugin.BLE.Abstractions;
 using Plugin.BLE.Abstractions.Contracts;
+using Plugin.BLE.BroadcastReceivers;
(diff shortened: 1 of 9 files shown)


janusw:
You may be aware that we have another open PR on this topic (#691). Could you comment on how this PR relates to the other one and how they differ?

jefffhaynes:
I was not aware, sorry. You can close this, I just needed bonding support.

janusw:
> I was not aware, sorry.

Nevermind ;)


> You can close this, I just needed bonding support.

You're not the only one with this demand, apparently :)

Could you maybe check if #691 works well for your purpose, or if you see any problems?

jefffhaynes:
I'll have to try it out later but glancing through it it looks like it should work. I'll switch over once the PR is merged. Thanks.

janusw:
> I'll have to try it out later but glancing through it it looks like it should work. I'll switch over once the PR is merged. Thanks.

Well, the other PR is not fully done yet, there are still some problems. If you have an interest in having this feature available, it would be good if you could try it now. Otherwise it might not get merged for 3.0 at all (or only partially).

Maybe your PR even works better than the other one? That should be checked, and I only have limited time for this right now.

Tobi-Tobi:
I've merged the changes from this branch into #691 and created a new PR #730 

This PR may be closed?

Issue #727, opened 2023-08-05
Add support for RequestMtuAsync on Windows


TravisRo:
## Steps to reproduce

1. Create MAUI App targeting WIndows with BLE.Plugin library

2. Call RequestMtuAsync

3. -1 is returned

## Expected behavior
It should return the MTU size that windows has already negotiated.  Windows always tries to negotiate the maximum MTU size.

## Actual behavior
Function does nothing and returns -1

### Crashlog

None

## Configuration

**Version of the Plugin: 3.0.0-beta.5

**Platform: Windows

**Device: ESP32 MINI-01

smsissuechecker:
Hi @TravisRo,

I'm the friendly issue checker.
It seems like (37.50 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

If you think it's fine to make an exception, just ignore this message.
**But if you think it was a mistake to delete the template, please close the issue and create a new one.**

Thanks!

janusw:
Fixed via #728.

PR #728, opened 2023-08-05
Add WIndows support for RequestMtuAsync. #727

TravisRo:


janusw (approved):
This looks pretty good to me, thanks for the contribution.

`MaxPduSize` sounded a bit strange to me at first sight, and I had to look up the documentation at https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattsession.maxpdusize,
which states:

> Gets the maximum protocol data unit (PDU) size. This metric is also known as the maximum transmission unit (MTU) size.

TravisRo:
Yeah, I don't know why they call it that.  I verified that it returns the same value for my ESP32 as Android (517)

Issue #497, opened 2021-03-22
BlueTooth Pairing Dialogue appers Twice


lrkPrecure:
have downloaded the Ble project and run it on Samsung galaxy s20+ against mobile debugging. Link of the project where we can see the issues https://github.com/xabre/xamarin-bluetooth-le Issues: "Bluetooth Pairing Dialogue appears Twice " ::::::::Not able to trace out where it is really occurring in second time, background "connect to device and pairing methods running on" I tried to do "Cancel" of "Bluetooth Pairing Dialogue" box, none of the intent.Action gets fired on "OnReceive" method. So it is difficult to fix when cancel button clicks. Please guide me ASAP. our project is almost on Prod.

## Steps to reproduce

1.Connect  and do pairing of the external device

2."Bluetooth Pairing Dialogue appears " :::::::twice continuously

3.  Not  able to see where it goes when we presses "Cancel" button

## Expected behavior
Tell us what should happen

## Actual behavior
Tell us what happens instead

### Crashlog

If something causes an exception paste full stack trace + Exception here

## Configuration

**Version of the Plugin:** e.g. 1.0.0 / 1.0.1-alpha3

**Platform:**  Android 10

**Device:**  Samsung galaxy s20+

smsissuechecker:
Hi @lrkPrecure,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

sutterv:
I have the same issue. I have to connect (pair) a protected device. Is it possible to manage the paring programmaticaly ?

Auto72:
I have Google Pixel 7 phone and when I connect for the first time with the BLE device (not yet bonded) it asks to Pairing two times (pop up window), one after the other. One time should be enough. The BLE device requires secure communication and has no PIN and no keyboard.

PR #730, opened 2023-08-09
Bonding on Android the Third

Tobi-Tobi:
As we're needing the bonding functionality and saw that there is help needed we decided to create our own pull request.

The error that [jeffhaynes](https://github.com/jefffhaynes) described in #724 also affects us => "bonding dialog appearing twice" also tracked by #497 

I've successfully tested that it doesn't occur anymore (Android 10 and 13). After testing around I found some bits missing from the #691 implementation. As #724 got written without the knowledge of #691 that is to be excepting.

So I sat down and merged the changes from #724 into #691.
 
The whole PR is tested with the following smartphones: 
1. Samsung S9 (Android 10)
2. Google Pixel 6A with latest Android 13


janusw:
> As we're needing the bonding functionality and saw that there is help needed we decided to create our own pull request.

Thanks for working on this. Bonding seems to be our most-demanded feature right now 😃, so I think it would be very nice if we could get this into 3.0.

I'll try to do a basic review soon, and I also hope that @midoragh and @jefffhaynes will be able to test and review this (or anyone else interested in this feature).

Furthermore, we'll need to find out why the CI build on Windows broke. It does not seem that the changes in this PR are to blame. Instead it looks it the error appeared when the .NET SDK on the build host changed from 7.0.306 to 7.0.400.

janusw:
> Furthermore, we'll need to find out why the CI build on Windows broke. It does not seem that the changes in this PR are to blame. Instead it looks it the error appeared when the .NET SDK on the build host changed from 7.0.306 to 7.0.400.

This is fixed via #737, but the CI build still fails here:

`BondStatusBroadcastReceiver.cs(37,33): error CS0117: 'OperatingSystem' does not contain a definition for 'IsAndroidVersionAtLeast' [D:\a\dotnet-bluetooth-le\dotnet-bluetooth-le\Source\Plugin.BLE\Plugin.BLE.csproj::TargetFramework=MonoAndroid10.0]`

@Tobi-Tobi Please fix!

Tobi-Tobi:
@janusw Problem solved 👍

janusw:
> @janusw Problem solved 👍

Alright, thanks for fixing!

Note that I will only merge this if @midoragh @jefffhaynes are happy with it. So guys, please speak up: Is this ok with you, or do you see any problems?

jefffhaynes:
Sorry, I've been traveling for the last two weeks and don't have the device I was using with me but it looks correct to me. I'll try it as soon as I can.

Source/Plugin.BLE/Android/BroadcastReceivers/BondStatusBroadcastReceiver.cs
```diff
@@ -1,41 +1,59 @@
...
 
-            switch (bondState)
+            BluetoothDevice bluetoothDevice;
+            
+            // As older versions of .NET (pre7) don't include the Tiramisu version code, handle it manually.
+#if NET7_0_OR_GREATER
+	        if (Build.VERSION.SdkInt >= BuildVersionCodes.Tiramisu)
+#else
+            if ((int)Build.VERSION.SdkInt >= 33)
+#endif
```
janusw, line 38: One comment here: The usual pattern for this kind of SDK version check in other parts of the library is ...

```
#if NET6_0_OR_GREATER
                    OperatingSystem.IsAndroidVersionAtLeast(33)
#else
                    (Build.VERSION.SdkInt >= BuildVersionCodes.Tiramisu)
-#endif
```

Is there a particular reason why you chose to do it differently, or just "trying to make CI work somehow"? If I understand your comment correctly, then the `Tiramisu` version code is not available everywhere (although I don't fully understand why). Maybe just use the number instead, but stick to the rest of the pattern?

I think the idea of the above pattern was to use the modern way of doing this on recent .NET versions (i.e. `OperatingSystem.IsAndroidVersionAtLeast`), and the legacy way on .NET Standard / Xamarin (i.e. `Build.VERSION.SdkInt`).

I would prefer to make all such checks in a consistent manner, unless there is a good reason to deviate from it.

Tobi-Tobi: I changed it to your implementation, yet again that throws the error that Tiramisu is not available, see the failed github action:  
`D:\a\dotnet-bluetooth-le\dotnet-bluetooth-le\Source\Plugin.BLE\Android\BroadcastReceivers\BondStatusBroadcastReceiver.cs(37,59): error CS0117: 'BuildVersionCodes' does not contain a definition for 'Tiramisu'`  
Is a cast alright in this case? 

I didn't know that `OperatingSystem.IsAndroidVersionAtLeast(xyz)` is the preferred way in newer versions, updated it accordingly.

janusw: > I changed it to your implementation, yet again that throws the error that Tiramisu is not available (which I tried to evade by my implementation), see the failed github action: `D:\a\dotnet-bluetooth-le\dotnet-bluetooth-le\Source\Plugin.BLE\Android\BroadcastReceivers\BondStatusBroadcastReceiver.cs(37,59): error CS0117: 'BuildVersionCodes' does not contain a definition for 'Tiramisu'` Is a cast alright in this case?

I think the failure is due to the fact that we're not targeting Android 13 yet. So, yes, casting to int is ok with me. Thanks for your adjustments!

Source/Plugin.BLE/Android/BroadcastReceivers/BondStatusBroadcastReceiver.cs
```diff
@@ -1,41 +1,59 @@
...
-                    break;
-
-                case Bond.Bonding:
-                    BondStateChanged(this, new DeviceBondStateChangedEventArgs() { Device = device, State = DeviceBondState.Bonding });
-                    break;
+#if NET7_0_OR_GREATER
+	            bluetoothDevice = (BluetoothDevice)intent.GetParcelableExtra(BluetoothDevice.ExtraDevice, Java.Lang.Class.FromType(typeof(BluetoothDevice)));
+#else
+	            bluetoothDevice = (BluetoothDevice)intent.GetParcelableExtra(BluetoothDevice.ExtraDevice);
+#endif
```
janusw, line 44: Another stupid question here: Why is the additional condition on NET7.0 necessary here? The variant with two arguments is only available on Android 33+ AFAICS, and you already have a check on the Android version above.

It is probably necessary because we are not targeting Android 13 in the classic Xamarin context, right? So maybe we should really update the targeted Android version for Xamarin. I can look into that.

janusw: > It is probably necessary because we are not targeting Android 13 in the classic Xamarin context, right? So maybe we should really update the targeted Android version for Xamarin. I can look into that.

see #740

Tobi-Tobi: image  
Older apps don't know the new overload added in api 33 right?

When using #740 changes and then removing the version check it runs into the following problem:
```
/Users/tobi/1_Code/3rd-Party/BT-Dotnet/dotnet-bluetooth-le-jeffhaynes/Source/Plugin.BLE/Android/BroadcastReceivers/BondStatusBroadcastReceiver.cs(40,56): error CS1501: No overload for method 'GetParcelableExtra' takes 2 arguments [/Users/tobi/1_Code/3rd-Party/BT-Dotnet/dotnet-bluetooth-le-jeffhaynes/Source/Plugin.BLE/Plugin.BLE.csproj::TargetFramework=net6.0-android]
```
I'm not exactly sure if we can remove it even when using your changes. As the error may be a problem of my toolchain I pushed the MonoAndroid update with the removed check as it's own commit here. If the pipeline passes everything is fine, if not I'll revert it.

Tobi-Tobi: Yeah.. The CI also crashed because of it:  
```
Error: /Users/runner/work/dotnet-bluetooth-le/dotnet-bluetooth-le/Source/Plugin.BLE/Android/BroadcastReceivers/BondStatusBroadcastReceiver.cs(40,56): 
error CS1501: No overload for method 'GetParcelableExtra' takes 2 arguments 
[/Users/runner/work/dotnet-bluetooth-le/dotnet-bluetooth-le/Source/Plugin.BLE/Plugin.BLE.csproj::TargetFramework=net6.0-android]
```

I reverted it. So the MonoAndroid update can be merged in its own branch.

janusw: Ok, very nice. Thanks.

Source/Plugin.BLE/Android/BroadcastReceivers/BondStatusBroadcastReceiver.cs
```diff
@@ -1,41 +1,55 @@
...
 
-            switch (bondState)
+            BluetoothDevice bluetoothDevice;
+            
+            // As older versions of .NET (pre7) don't include the Tiramisu version code, handle it manually.
+#if NET6_0_OR_GREATER
+	        if (OperatingSystem.IsAndroidVersionAtLeast(33))
+#else
+            if ((int)Build.VERSION.SdkInt >= 33)
+#endif
```
janusw, line 38: With #740 merged, you should also be able to use the Tiramisu version code here (instead of casting to int). Then the comment in line 33 can go away.
Also: The whitespace/indentation in line 35 looks odd.

Source/Plugin.BLE/Android/BroadcastReceivers/BondStatusBroadcastReceiver.cs
janusw: I re-triggered the CI and it succeeded now. 🎉

Source/Plugin.BLE/Android/BroadcastReceivers/BondStatusBroadcastReceiver.cs
Tobi-Tobi: Ah, somehow a Tab sneaked in. I blame Rider 😄 

Changed the cast to BuildVersionCode.Tiramisu.   
Also fixed the indentation 🙈

janusw: Awesome. I think this block of code is finally where it should be. Thanks 😁

janusw:
So, originally I wanted to have a release 3.0 in August, now we're a bit late already. This PR is really the last bigger feature I'm willing to consider for the final release.

But since it combines elements of two earlier PRs for the same feature, I will not merge it unless the authors of the earlier PRs (@midoragh and @jefffhaynes) signal their approval. It hope that they could either review the code or do some testing (ideally both).

If we don't get any feedback by the end of September, I will publish the final release 3.0 without this feature (which will then have to wait for 3.1). I really hope we can still get it into 3.0, but I don't want to delay the release too much.

jefffhaynes:
I'll test tonight. Thanks.

jefffhaynes:
Initially had some super weird behavior but it seems like Android just didn't want to fully forget the previously bonded devices so I had to reset network settings (Android ftw).  However, after that everything worked smoothly.

Looks good, thanks for the update.

janusw:
> Looks good, thanks for the update.

Thanks for the positive feedback, @jefffhaynes !

@midoragh Do you think you'll be able to do some code review and/or testing in the next two weeks?

janusw (approved):

janusw:
It looks like we're not getting any comment from @midoragh here, but we have an approval from @jefffhaynes and all my comments have been addressed, so I will simply merge this now (and tag an rc for 3.0). Thanks for the contribution, @Tobi-Tobi !

janusw:
I prepared the tag and uplodaded to nuget. If any problems come up (or we get a late review), we can still fix it for the final version 3.0, which I would like to release in about a week from now (if no blockers appear in the meantime).

Tobi-Tobi:
> I prepared the tag and uplodaded to nuget. If any problems come up (or we get a late review), we can still fix it for the final version 3.0, which I would like to release in about a week from now (if no blockers appear in the meantime).

Thank you @janusw  
We're also going to update to the RC soon (early next week). If any problems arise I'll create an issue or fix it. 👍

Tobi-Tobi:
Update: 

We tested it some minutes ago, still works as expected with the RC. 👍

Issue #734, opened 2023-08-11
DisconnectAsync hangs on Windows and Android after scanning then connecting with ConnectToKnownDeviceAsync


TravisRo:
To help us fix your issue, please provide the information in the below template. If something causes a crash, provide as much information as you can gather.
Just imagine: we do not know what you are doing! 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.

## Steps to reproduce

1. Start device scanning

2. Stop device scanning after test device is discovered

3. Connect to test device using `ConnectToKnownDeviceAsync`

4. Disconnect from test device

5. Test device does not disconnect and `DisconnectDeviceAsync` hangs

## Expected behavior
Device should disconnect successfully

## Actual behavior
Device does not disconnect and `DisconnectDeviceAsync` hangs

### Crashlog

If something causes an exception paste full stack trace + Exception here

## Configuration

**Version of the Plugin:** 3.0.0-beta.5

**Platform:** Android 13/Windows 10

**Device:** Galaxy S23 Ultra/Windows 10

## Sample code
```CSharp
using System.Collections.ObjectModel;
using MetroLog;
using Plugin.BLE;
using Plugin.BLE.Abstractions;
using Plugin.BLE.Abstractions.Contracts;
using Plugin.BLE.Abstractions.EventArgs;

namespace MauBleTest
{
    public class DeviceItem
    {
        public string Name { get; set; }
        public IDevice Device { get; set; }
    }
    public class MainViewModel:TinyMvvm.TinyViewModel
    {
        internal static readonly Guid TTS_BLE_SERVICE_ID = Guid.Parse("{0000abf0-0000-1000-8000-00805f9b34fb}");
        internal static readonly Guid TTS_BLE_CHAR_DATA_RECEIVE_ID = Guid.Parse("{0000abf1-0000-1000-8000-00805f9b34fb}");
        internal static readonly Guid TTS_BLE_CHAR_DATA_NOTIFY_ID = Guid.Parse("{0000abf2-0000-1000-8000-00805f9b34fb}");
        internal static readonly Guid TTS_BLE_CHAR_COMMAND_RECEIVE_ID = Guid.Parse("{0000abf3-0000-1000-8000-00805f9b34fb}");
[...]

PR #735, opened 2023-08-11
Fix  DisconnectAsync hangs on Windows and Android after scanning then…

TravisRo:
… connecting with ConnectToKnownDeviceAsync #734

In the native Windows and Android adapter, it always create a new native `IDevice` instance when using `ConnectToKnownDeviceAsync` if the device has already been discovered via scanning you end up with two `IDevice` instances.  If the user then calls `DisconnectDeviceAsync` on the `IDevice` instance that was discovered via scanning, this issue occurs.

I DID NOT check the other platform adapters so this issue may be present in other platforms.  The solution is easy; just check the `DiscoveredDevicesRegistry` before creating a new `Device` instance.

### DIFF
--- a/Source/Plugin.BLE/Android/Adapter.cs
+++ b/Source/Plugin.BLE/Android/Adapter.cs
@@ -242,11 +242,12 @@ namespace Plugin.BLE.Android
             ((Device)device).Disconnect();
         }
 
-        public override async Task<IDevice> ConnectToKnownDeviceAsync(Guid deviceGuid, ConnectParameters connectParameters = default(ConnectParameters), CancellationToken cancellationToken = default(CancellationToken))
+        public override async Task<IDevice> ConnectToKnownDeviceNativeAsync(Guid deviceGuid, ConnectParameters connectParameters = default(ConnectParameters), CancellationToken cancellationToken = default(CancellationToken))
         {
             var macBytes = deviceGuid.ToByteArray().Skip(10).Take(6).ToArray();
             var nativeDevice = _bluetoothAdapter.GetRemoteDevice(macBytes);
-
+            if (nativeDevice == null)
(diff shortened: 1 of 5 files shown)


travis012:
Ohh no!  the builder must be broken..

wasn't me.. :)

janusw:
> Ohh no! the builder must be broken..
> 
> wasn't me.. :)

Yeah, it certainly does not look like you caused it. It happens in another PR as well. Might be due to updated tooling versions on the build host. Will try to look into that ...

janusw:
Alright, I fixed the CI issues in #737 and rebased this one, so that all build checks succeed now.

janusw (changes requested):

TravisRo:
Yeah that sounds good to me.  I can handle that.  I don't have a way to test on iOS and MAC so someone should probably look it over.  I also think we should add our device object to the `DiscoveredDevicesRegistry` if it's not already in there. Provided of-course that `ConnectToKnownDeviceAsync` succeeds.  This will cover us on both ends.

janusw:
> Yeah that sounds good to me. I can handle that.

Awesome, thanks! :+1:

> I don't have a way to test on iOS and MAC so someone should probably look it over.

I can possibly take care of that part. In any case, if your generic code works well on Android and Windows, I think the probability is quite low that it will cause problems on iOS (given that it is reasonably simple).

> I also think we should add our device object to the `DiscoveredDevicesRegistry` if it's not already in there. Provided of-course that `ConnectToKnownDeviceAsync` succeeds. This will cover us on both ends.

Yes, I think that makes sense.

TravisRo:
Previous changes reverted and new proposed changes pushed.  Tested on Windows and Android.  I've also updated code to throw a `DeviceConnectionException` as apposed to a generic `Exception` when `ConnectToKnownDeviceNativeAsync` cannot find the device being requested.

janusw (approved):
LGTM now. Thanks a lot for the contribution!

PR #736, opened 2023-08-19
Maui client

IanBUK:
This is a sample Maui client, I've run it on MacOS and a Samsung A8 Android tablet.

It simply scans for BLE items. The app doesn't do connections etc.


IanBUK:
I've not run this locally on a Windows box - I don't have one! The M1 MacBook Air was just too tempting. Sometimes I wonder whether it was a good move - I miss some of the nice tools available only on Windows.

IanBUK:
One reason why the Mac build might fail here whilst it works on my machine is that I'm building the BLE.Mac.Maui.sln

Tobi-Tobi:
For the Mac Build you want to create the NuGet via Bash or ZSH I guess.  
That is the way the CI handles it and will result in the same errors when trying to create the NuGet on your mac.

```bash
WORKDIR=$(git rev-parse --show-toplevel)
FINAL_VERSION=$(git describe)

echo "📦 Build Plugin.BLE"
dotnet build $WORKDIR/Source/Plugin.BLE/Plugin.BLE.csproj /p:Configuration=Release /t:restore,build,pack /p:Version=$FINAL_VERSION /p:ContinuousIntegrationBuild=true /p:DeterministicSourcePaths=false
```

janusw:
> The windows `Action` is broken on other requests as well but should also run into the same problematic later on when the general windows runner build is fixed.

The general CI problems (on Windows) are fixed via #737, and I rebased this PR, but still the builds on Windows and Mac continue to fail with the same error:

`Error: /Users/runner/work/dotnet-bluetooth-le/dotnet-bluetooth-le/Source/MvvmCross.Plugins.BLE/Plugin.shared.cs(6,17): error CS0234: The type or namespace name 'Extensions' does not exist in the namespace 'Microsoft' (are you missing an assembly reference?) [/Users/runner/work/dotnet-bluetooth-le/dotnet-bluetooth-le/Source/MvvmCross.Plugins.BLE/MvvmCross.Plugins.BLE.csproj::TargetFramework=net7.0]`

The `using Microsoft.Extensions.Logging;` in line 6 of `Plugin.shared.cs` was present already before this PR (where it did not cause any error), so it seems the error is due to the addition of the target framework `net7.0` to `MvvmCross.Plugins.BLE.csproj`. @IanBUK Why was this addition necessary? Could it be removed again?

If not, one should probably just add a package reference to `Microsoft.Extensions.Logging.Debug` to `MvvmCross.Plugins.BLE.csproj`.

IanBUK:
net7.0 was added when I was trying to get it all working on my Mac, I've removed the reference and have added the nuget as suggested.

Source/BLE.Mac.Maui.sln
janusw, line 1: Actually I don't think we need another solution here. Instead, the project `BLE.Client.Maui` should be added to the existing `BLE.sln` (which contains everything) as well as `BLE.Mac.slnf` (which excludes the Xamarin client).

janusw (changes requested):

Source/BLE.Mac.Maui.sln
janusw: Ok, it seems you have added the new project to the existing sln/slnf files. Thanks for that.

IMHO the new `BLE.Mac.Maui.sln` that you are adding is unnecessary and should be removed again. Or is there any reason why you'd want to keep it?

IanBUK:
I kept it as my code didn’t reference the MvvM projects. It was as lightweight as I could get it.

janusw:
> I kept it as my code didn’t reference the MvvM projects. It was as lightweight as I could get it.

Ah, ok. Note that `BLE.Mac.slnf` is already a bit more leightweight than `BLE.sln`. Couldn't you just use that one? It contains a bit more than is needed for the MAUI sample, but OTOH it doesn't take incredibly long to load etc, right?

If you really insist that you need an additional sln on top of what we already have, please use a solution filter (slnf), which is more leightweight than a full-blown solution file and does not duplicate all of the configuration mapping etc. To do that, you could just copy `BLE.Mac.slnf` and remove all the stuff you don't want (this could then be called `BLE.Maui.Client.slnf` or similar).

janusw:
I found some time to test the client by now, with the following results:
* On MacOS 13.5, the scanning seems to work and finds several devices, but it's almost impossible to read the results (white font on yellow background!).
* On Android 13, I can start the scan, but it looks like no devices are being discovered at all.
* On iOS 16.6, the app crashes instantly ("Got a SIGABRT while executing native code. This usually indicates a fatal error in the mono runtime or one of the native libraries used by your application.")

@IanBUK Does it work on Android for you? If yes, which version of Android? Have you tried iOS?

IanBUK:
One thought about Android - check the permissions for the app on the device. I’ve had ‘fun’ with this for Godot apps. It seems to need ‘Location’ and ‘Nearby Devices’.On 29 Aug 2023, at 21:38, Janus Weil ***@***.***> wrote:﻿
I found some time to test the client by now, with the following results:

On MacOS 13.5, the scanning seems to work and finds several devices, but it's almost impossible to read the results (white font on yellow background!).
On Android 13, I can start the scan, but it looks like no devices are being discovered at all.
On iOS 16.6, the app crashes instantly ("Got a SIGABRT while executing native code. This usually indicates a fatal error in the mono runtime or one of the native libraries used by your application.")

@IanBUK Does it work on Android for you? If yes, which version of Android? Have you tried iOS?

Source/BLE.Mac.Maui.sln
IanBUK: Removed the new solution file.

janusw:
> **MacOS 13.5** The text appears black on my screen - it must be a theming issue. I've changed it to use a ListView.ItemTemplate based around a grid, with hard coded colours - black text on a light grey background.

Yes, looks much better now. Thank you!

janusw (approved):
@IanBUK Your latest commit fixes the client on Android, and I added two more commits that make it work on iOS as well.

With that, this PR is basically ok for me. Will merge soon.

(I assume it will not work on Windows yet, but I cannot test that right now, and it can still be fixed later ...)

janusw:
Functionally, one can think of a good number of improvements and extensions as well, but I think the current state is good enough as a simple sample client, and should serve as a starting point for further improvements.

So, yes, I'm merging now. Thanks for the contribution!

PR #746, opened 2023-09-20
Windows.Devices.Bluetooth Only

AskBojesen:
Modifying the Windows implementation of the Plugin NOT to use `CommunityToolkit.WinUI.Connectivity.BluetoothLEHelper` and the `CommunityToolkit.WinUI.Connectivity.ObservableBluetoothLEDevice` that are only working if a WINUI `Microsoft.UI.Dispatching.DispatcherQueue` is present.
This means we can remove dependencies to Microsoft.Toolkit.Uwp.Connectivity and CommunityToolkit.WinUI.Connectivity nuget packages.

Only use `Windows.Devices.Bluetooth.*` API

I have added a BLE.Client.WinConsole demo in order to show the simple functionality

I have tested that the BLE.Client.UWP is still working


janusw:
Thanks for the contribution, I'll try to have a look at this soon.

I'm afraid it's slightly late to be included in 3.0, but I will surely consider it for a version 3.1.

AskBojesen:
> Thanks for the contribution, I'll try to have a look at this soon.
> 
> I'm afraid it's slightly late to be included in 3.0, but I will surely consider it for a version 3.1.

Thanks! No stress we are simply compiling the code for our Windows application using the same API as for the Apple and Android until then.

Should I just let this pull request be there waiting for some 3.1 in the future?

janusw:
> Thanks! No stress we are simply compiling the code for our Windows application using the same API as for the Apple and Android until then.

:+1:

> Should I just let this pull request be there waiting for some 3.1 in the future?

Yes, please. I will try to have a look at it soon. We can hopefully merge it after 3.0 is released.

HarmenSchouten:
For a project I have been using the source branch of this PR to get my devices, because it works way better than the original implementation for a .NET MAUI application. 

Now in my case I have to get the `SystemConnectedDevices`, since this isn't implemented yet, I wrote something for it. It's certainly less than ideal because it's gotta be async while the method isn't async and you have to do the `.GetAwaiter().GetResult()` dance.
Is it at all possible to include an implementation for this function, I'd submit it myself if necessary, to this PR?

AskBojesen:
_> For a project I have been using the source branch of this PR to get my devices, because it works way better than the original implementation for a .NET MAUI application._ 
Thanks! I agree - the new implementation is much better - I use it everyday in our project.
@janusw: What is the process now to get this PR into master?
That would also ease my daily work not to compile my own version.

_> Now in my case I have to get the `SystemConnectedDevices`, since this isn't implemented yet for Windows, I wrote something for it. It's certainly less than ideal because it's gotta be async while the method isn't async and you have to do the `.GetAwaiter().GetResult()` dance. Is it at all possible to include an implementation for this function, I'd submit it myself if necessary, to this PR?_

Yes, I see that `IReadOnlyList<IDevice> GetSystemConnectedOrPairedDevices(Guid[] services = null)` is not implemented in Windows - I can implement in this branch also. Maybe you @HarmenSchouten can help me understand a use case for this method, since we do not need it in our project. And likewise for the GetBondedDevices if you need that also - this is not implemented for WIndows either.

janusw:
> For a project I have been using the source branch of this PR to get my devices, because it works way better than the original implementation for a .NET MAUI application.

Can you explain in which ways it works better? What are the differences?


> Now in my case I have to get the `SystemConnectedDevices`, since this isn't implemented yet for Windows, I wrote something for it. It's certainly less than ideal because it's gotta be async while the method isn't async and you have to do the `.GetAwaiter().GetResult()` dance. Is it at all possible to include an implementation for this function, I'd submit it myself if necessary, to this PR?

Well, you cannot include it directly in this PR. For now, you could set up a branch in your fork on top of @AskBojesen's branch and push the implementation there. Then you could either wait until this PR merged and then create your own PR for the additional features, or @AskBojesen could include your changes in his branch (and thus, in this PR).

janusw:
> Thanks! I agree - the new implementation is much better - I use it everyday in our project. @janusw: What is the process now to get this PR into master?

Well, it should be reviewed and tested. I just verified that the MAUI sample client works well on this branch. At least it is able to scan and discover devices.

> That would also ease my daily work not to compile my own version.

Note: You shouldn't have to compile it yourself. For each PR, the GitHub Actions CI builds nupkgs that you can use.


I'll try to do a basic code review soon (maybe I will manage over the weekend). It would be good if some users (who are interested in the Windows implementation) could test this branch as well and report their findings.

HarmenSchouten:
> Maybe you @HarmenSchouten can help me understand a use case for this method, since we do not need it in our project.

The use case is that we're writing something which will send colour signals to a peripheral device, like a mouse or keyboard. Ideally the user connects the device by themselves to the PC, leaving only the software to find the device and send the command when necessary. With regards to the BondedDevices, this is not something we are using either, so I'll leave that open for now.

> Can you explain in which ways it works better? What are the differences?

When I got to play around with the package I tried to get the ConnectedDevices, which ended up working as expected. However, once I got there and just tried to get the application running with a service in the background responsible for getting the required devices, I kept crashing on an error coming from the package. IIRC it had something to do with a QueueDispatcher. It's been a couple weeks since I've tried, I can maybe deliver a more detailed report later this weekend.

> Well, you cannot include it directly in this PR. For now, you could set up a branch in your fork on top of @AskBojesen's branch and push the implementation there. Then you could either wait until this PR merged and then create your own PR for the additional features, or @AskBojesen could include your changes in his branch (and thus, in this PR).

As I expected, I'll make sure to create a PR to @AskBojesen' branch with the implementation.

AskBojesen:
@HarmenSchouten: Looking forward to see your code for GetSystemConnectedOrPairedDevices in Windows
I have attached I I would implement it - to go into Adapter.cs

[GetSystemConnectedOrPairedDevices.txt](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/files/13195692/GetSystemConnectedOrPairedDevices.txt)

Source/Plugin.BLE/Windows/Extensions/GattProtocolErrorParser.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        if (errorValue == GattProtocolError.UnsupportedGroupType)
+        {
+            return "Unsupported Group Type";
+        }
+
+        if (errorValue == GattProtocolError.WriteNotPermitted)
+        {
+            return "Write Not Permitted";
+        }
```
janusw, line 120: IMHO this whole block here should better be one large `switch` statement than a long succession of `if` statements.

Source/Plugin.BLE/Windows/Extensions/GattResultExtensions.cs
```diff
@@ -48,15 +43,15 @@ public static void ThrowIfError(this GattCommunicationStatus status, [CallerMemb
         }
 
         private static string GetErrorMessage(this GattCommunicationStatus status, string tag, byte? protocolError)
-        {
+        {
```
janusw, line 46: extra whitespace here

janusw (commented):
Skimming over the code changes, it all looks pretty good to me. It's nice that the previous division into UWP and WinUI implementation is removed, the code is unified and many `#if`s are removed. I only have two minor comments.

AskBojesen:
> I'll try to do a basic code review soon (maybe I will manage over the weekend). It would be good if some users (who are interested in the Windows implementation) could test this branch as well and report their findings.

@janusw : I have three concerns myself regarding the implementation I have made.

1. `namespace Plugin.BLE.UWP` It is not only UWP now - it is for all Windows. (netstandard + .Net + UWP)
But having a namespace namespace Plugin.BLE.Windows will cause name collision with Windows
Maybe Plugin.BLE.Win instead to show that it's not only UWP when reading code ?

2. The methods in  `Plugin.BLE\Windows\Extensions\TextHelpers.cs` - They are nice for logging and so, but we could share such method in the Shared folder - there are some variants and copies in the Apple and Android folders

3. There is a mix now between having Id's as `Guid` or as `string` 
This causes a lot of conversion to/from `Guid` from/to `string`
This is of course not introduced in this PR, but I would like to know why it is this mix.
Eg. in the `Source\Plugin.BLE\Shared\AdapterBase.cs`:
 protected ConcurrentDictionary<Guid, IDevice> DiscoveredDevicesRegistry { get; } = new ConcurrentDictionary<Guid, IDevice>();
 public ConcurrentDictionary<string, IDevice> ConnectedDeviceRegistry { get; } = new ConcurrentDictionary<string, IDevice>();`
 And in the `Plugin.BLE\Windows\Adapter.cs`:
`private readonly IDictionary<string, IDevice> _deviceOperationRegistry = new ConcurrentDictionary<string, IDevice>();
`

AskBojesen:
@janusw : Could we have a teams call - if you sent an email to me ab@egoo.health - the I can sent an invite for call - or something else?

janusw:
> It would be good if some users (who are interested in the Windows implementation) could test this branch as well and report their findings.

☝️
@TravisRo @Praseodymi @DamianSuess @frankfolsche @fozzzgate @vitalii-vov
Any comments from your side on this PR?

Source/Plugin.BLE/Windows/Extensions/GattProtocolErrorParser.cs
AskBojesen: I will rather not change this file, since it is  this file:
https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/main/Microsoft.Toolkit.Uwp.Connectivity/BluetoothLEHelper/GattProtocolErrorParser.cs
I have now added comments in the top of the file to clarify this.
The reason for not just including the `CommunityToolkit.WinUI.Connectivity` package is that this `GetErrorString` method is the only component/class, which is needed and the CommunityToolkit.WinUI.Connectivity is a monster IMO

janusw:
> > I'll try to do a basic code review soon (maybe I will manage over the weekend). It would be good if some users (who are interested in the Windows implementation) could test this branch as well and report their findings.
> 
> @janusw : I have three concerns myself regarding the implementation I have made.
> 
>     1. `namespace Plugin.BLE.UWP` It is not only UWP now - it is for all Windows. (netstandard + .Net + UWP)
>        But having a namespace namespace Plugin.BLE.Windows will cause name collision with Windows
>        Maybe Plugin.BLE.Win instead to show that it's not only UWP when reading code ?

Right, I missed this point. The UWP namespace is somewhat historic, and already before this PR was used for WinUI as well. So yes, I think it should be changed.

I would actually go with `Plugin.BLE.Windows`. That also matches the folder structure in the project. Are you concerned that this would create conflicts with the global Windows.* namespaces? I don't think that would be a real problem (and we have such a case already with Plugin.BLE.Android and global::Android.*).

Gladly a namespace change in the platform-specific part should not lead to any source incompatibility in user code (because users typically only reference the generic interfaces in the namespaces `Plugin.BLE` and `Plugin.BLE.Abstractions`).


>     2. The methods in  `Plugin.BLE\Windows\Extensions\TextHelpers.cs` - They are nice for logging and so, but we could share such method in the Shared folder - there are some variants and copies in the Apple and Android folders
> 
>     3. There is a mix now between having Id's as `Guid` or as `string`
>        This causes a lot of conversion to/from `Guid` from/to `string`
>        This is of course not introduced in this PR, but I would like to know why it is this mix.
>        Eg. in the `Source\Plugin.BLE\Shared\AdapterBase.cs`:
>        protected ConcurrentDictionary<Guid, IDevice> DiscoveredDevicesRegistry { get; } = new ConcurrentDictionary<Guid, IDevice>();
>        public ConcurrentDictionary<string, IDevice> ConnectedDeviceRegistry { get; } = new ConcurrentDictionary<string, IDevice>();` And in the`Plugin.BLE\Windows\Adapter.cs`: `private readonly IDictionary<string, IDevice> _deviceOperationRegistry = new ConcurrentDictionary<string, IDevice>();
>        `

These are both good points, but IMHO both of them could be cleaned up in follow-up PRs.

About (3): I'm not sure where this mix comes from. Might just be an accidental inconsistency (e.g. code from different authors), or there might possibly be an actual reason for it (that I don't see right now).

Praseodymi:
I'll take a look at the code if I get the time, but I can say it's dropped into our project seamlessly, and it means the Windows version now replicates our Android build with no additional dependencies (also including my PR #764, obviously). It touched a lot of files to do it, but I think it's definitely a step in the right direction.

janusw:
> I'll take a look at the code if I get the time, but I can say it's dropped into our project seamlessly, and it means the Windows version now replicates our Android build with no additional dependencies (also including my PR #764, obviously). It touched a lot of files to do it, but I think it's definitely a step in the right direction.

Awesome, thanks for the comment 👍

AskBojesen:
> I would actually go with `Plugin.BLE.Windows`. That also matches the folder structure in the project. Are you concerned that this would create conflicts with the global Windows.* namespaces? I don't think that would be a real problem (and we have such a case already with Plugin.BLE.Android and global::Android.*).

Ok - I have changed namespace to `Plugin.BLE.Windows` and I have now learned the `global::` feature in C#, which I have not needed before.

Source/Plugin.BLE/Windows/Extensions/GattProtocolErrorParser.cs
janusw: Ok then, fine with me.

HarmenSchouten:
> @HarmenSchouten: Looking forward to see your code for GetSystemConnectedOrPairedDevices in Windows I have attached how I would implement it - to go into Adapter.cs
> 
> [GetSystemConnectedOrPairedDevices.txt](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/files/13195692/GetSystemConnectedOrPairedDevices.txt)

This implementation is almost identical to the one I had, in fact, it's a bit more complete with the logging and such. So if it's fine by you I'd vote to include that piece in this PR for the sake of completeness. No other remarks from my side.

AskBojesen:
> This implementation is almost identical to the one I had, in fact, it's a bit more complete with the logging and such. So if it's fine by you I'd vote to include that piece in this PR for the sake of completeness. No other remarks from my side.

Janus gave thumbs up - so now it is included in this PR

janusw:
> Janus gave thumbs up - so now it is included in this PR

Thanks. I just squashing the last two commits, in order to get rid of the fixup commit (I'm a fan of a clean commit history 😉).

janusw (approved):
This is ok with me by now.

janusw:
Merged. Thanks for the contribution, @AskBojesen!

PR #764, opened 2023-10-25
ReadAsync updates characteristic value on Windows

Praseodymi:
On Android calling ReadAsync on a Characteristic also updates it's Value. However this doesn't occur in Windows builds. The name of the local variable leads me to think this was a typo and updating the Value is the intended behaviour.


janusw:
> On Android calling ReadAsync on a Characteristic also updates it's Value. However this doesn't occur in Windows builds. The name of the local variable leads me to think this was a typo and updating the Value is the intended behaviour.

Yes, I agree. This seems to be a mistake. Apparently it has been introduced in commit 074801c35126bce3c68b1c1278e33842bfa0bcbc as part of #669, which has been merged between 3.0.0-beta.3 and 3.0.0-beta.4.

janusw (approved):
LGTM. Thanks for catching this!

PR #774, opened 2023-11-08
disable MainThreadInvoker in Adapter.ConnectToDeviceAsync

### LINKED ISSUE #757: After upgrading from 2.1.3 to 3, ConnectToKnownDeviceAsync became a single-device connectivity function with some odd observations
Everything was fine with 2.1.3 on Android until we had to upgrade to 3 in order to get our UWP version of a Xamarin forms app also to build and run. 

Then a range of oddities started to appear and caused a range of issues for us in production.

Most importantly, we have a range of BLE peripherals like printers and sensors that are all initialized at startup. We use ConnectToKnownDeviceAsync "without" any CancellationTokens and everything was fine with connectivity and auto-reconnect on 2.1.3 and it still is! We run multiple independent async tasks to try to connect to devices at startup. If a device is absent, the next ones will all be found consistently and are all initialized. When the absent device becomes available, we even observe that it is getting connected without an issue. Even when we disconnect them after connection, and re-connect them, still they become connected and all is healthy.

To our amazement, after upgrading to 3.0.0, we accidentally realized that in the absence of one device, ConnectToKnownDeviceAsync in one async function affects the behavior of ConnectToKnownDeviceAsync for another async function. If a device is unavailable, ConnectToKnownDeviceAsync  waits indefeintely for it to become available as expected but until the first call succeeds or fails, the next call to ConnectToKnownDeviceAsync never gets a chance to initialize and connect to another device, irrespectively, unlike v 2.1.3.

Now, with our religious attempt to force it to work with v3 and a timed CalcellationToken that we never had to use, this time we had to re-invent the wheel of running a re-try timer to try to connect to the absent devices so that upon the failure of the first one, then we have a chance to connect to the next one.

Result: it never worked. When the first device's ConnectToKnownDeviceAsync  fails on the timed CalcellationToken, then it affects the connectivity of the 2nd device that is present as if, whenever we do, the plugin at v3 can only work with one and only one device. When the first device is available and the 2nd one is unavailable, we can connect to the first one, but then when the retry timer for the 2nd device makes an attempt with ConnectToKnownDeviceAsync, then the first successful connection gets affected with dropouts and failures.

Let us be clear here: We know our onions very well and we're not fooling around and we really are time poor and we want to help. Hope you understand the mere fact of creating this ticket is a huge undertaking for us. So simply follow the instructions to recreate the situation instead of saying you can;t recreate the issue.

## Steps to reproduce

1. Use v 2.1.3 on Android, in a setup where you have more than 1 BLE device and you are initializing them all at one stage using ConnectToKnownDeviceAsync using multiple independent async functions, connect to them and observe healthy connectivity with read and write operations 

2. upgrade to v3

3. Make the first device in the queue unavailable and observe the behavior of ConnectToKnownDeviceAsync for the second one and so forth. Confirm that the 2nd and 3rd etc... calls to ConnectToKnownDeviceAsync do not succeed and remain dependent on the first call, no matter the setting.

## Expected behavior
Just like v 2.1.3, ConnectToKnownDeviceAsync must remain indefinitely waiting for the device to become available but when called on multiple async functions, it should make an attempt to connect to other devices irrespective of the previous outcomes. In other words, we should not have a single-device model for connections.

## Actual behavior
If the first call to ConnectToKnownDeviceAsync becomes stuck on an unavailable device, one can never connect to another device no matter the setting. If a timed CancellationToken is passed, the call to the first ConnectToKnownDeviceAsync does raise and exception, but after connecting to the next device, when trying to connect to the first device after a timer tick, the connection to the successfully-connected device also gets affected.

### Crashlog

If something causes an exception paste full stack trace + Exception here

## Configuration

**Version of the Plugin:** e.g. 3.0.0
**Platform:** e.g. Android 13 
**Device:** e.g. Samsung G525F.

janusw:
This is an attempt to fix #757. @skha83, could you please try this and tell me if it fixes the problem for you?

### DIFF
--- a/Source/Plugin.BLE/Shared/AdapterBase.cs
+++ b/Source/Plugin.BLE/Shared/AdapterBase.cs
@@ -218,7 +218,7 @@ namespace Plugin.BLE.Abstractions
 
                     subscribeReject: handler => DeviceConnectionError += handler,
                     unsubscribeReject: handler => DeviceConnectionError -= handler,
-                    token: cts.Token);
+                    token: cts.Token, mainthread: false);
             }
         }
 
--- a/Source/Plugin.BLE/Shared/Utils/TaskBuilder.cs
+++ b/Source/Plugin.BLE/Shared/Utils/TaskBuilder.cs
@@ -35,7 +35,8 @@ namespace Plugin.BLE.Abstractions.Utils
             Func<Action<Exception>, TRejectHandler> getRejectHandler,
             Action<TRejectHandler> subscribeReject,
             Action<TRejectHandler> unsubscribeReject,
-            CancellationToken token = default)
+            CancellationToken token = default,
+            bool mainthread = true)
         {
             var tcs = new TaskCompletionSource<TReturn>();
             void Complete(TReturn args) => tcs.TrySetResult(args);
@@ -51,7 +52,7 @@ namespace Plugin.BLE.Abstractions.Utils
                 subscribeReject(rejectHandler);
                 using (token.Register(() => tcs.TrySetCanceled(), false))
                 {
-                    return await SafeEnqueueAndExecute(execute, token, tcs);
+                    return await SafeEnqueueAndExecute(execute, token, tcs, mainthread);
                 }
             }
             finally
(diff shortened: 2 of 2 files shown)


janusw:
It was reported that this PR works well (see. https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/issues/757#issuecomment-1803320893), so I will merge it. Thanks for the feedback, @skha83.

RuT81:
I will be very happy if this can be fixed as soon as possible. I spent two days solving the same problem with multiple devices barcode scanner and printer. The barcode scanner usually goes to sleep to last as long as possible.
Disabling MainThreadInvoker helped resolve the situation.

PR #776, opened 2023-11-10
Improve Connect In Windows

AskBojesen:
In the Windows implementation, the ConnectToDeviceNativeAsync method did not work properly.
It was especially seen in the case when:

1. Connect
2. Get services
3. Disconnect
4. Connect
5. Get services

This has now been solved

### DIFF
--- a/Source/Plugin.BLE/Windows/Adapter.cs
+++ b/Source/Plugin.BLE/Windows/Adapter.cs
@@ -11,7 +11,6 @@ using Plugin.BLE.Abstractions.Contracts;
 using Plugin.BLE.Extensions;
 using System.Collections.Concurrent;
 using Windows.Devices.Enumeration;
-using WBluetooth = global::Windows.Devices.Bluetooth;
 
 namespace Plugin.BLE.Windows
 {
@@ -83,22 +82,6 @@ namespace Plugin.BLE.Windows
             nativeDevice.ConnectionStatusChanged -= Device_ConnectionStatusChanged;
             nativeDevice.ConnectionStatusChanged += Device_ConnectionStatusChanged;
 
-            // Calling the GetGattServicesAsync on the BluetoothLEDevice with uncached property causes the device to connect
-            //
-            // ref https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothledevice.frombluetoothaddressasync
-            // Creating a BluetoothLEDevice object by calling this method alone doesn't (necessarily) initiate a connection.
-            // To initiate a connection, set GattSession.MaintainConnection to true, or call an uncached service discovery
-            // method on BluetoothLEDevice, or perform a read/write operation against the device.
-            //
-            // Remark from Ask Bojesen 2023-11-12:
-            // Below three lines reflect the different approach with GattSession, but I could not get this working properly 
-            //
-            // var deviceId = BluetoothDeviceId.FromId(dev.NativeDevice.DeviceId);            
-            // var genericProfileGattSession = await GattSession.FromDeviceIdAsync(deviceId);            
-            // bool success = genericProfileGattSession.MaintainConnection = true;
-
-            //var servicesResult = await dev.NativeDevice.GetGattServicesAsync(BluetoothCacheMode.Uncached);
-            //bool success = servicesResult.Status == WBluetooth.GenericAttributeProfile.GattCommunicationStatus.Success;
             bool success = await dev.ConnectInternal(connectParameters, cancellationToken);
 
             if (!success)            
--- a/Source/Plugin.BLE/Windows/Adapter.cs
+++ b/Source/Plugin.BLE/Windows/Adapter.cs
@@ -84,7 +84,7 @@ namespace Plugin.BLE.Windows
 
             bool success = await dev.ConnectInternal(connectParameters, cancellationToken);
 
-            if (!success)            
+            if (!success)
             {
                 // use DisconnectDeviceNative to clean up resources otherwise windows won't disconnect the device
                 // after a subsequent successful connection (#528, #536, #423)
@@ -149,7 +149,7 @@ namespace Plugin.BLE.Windows
         protected override void DisconnectDeviceNative(IDevice device)
         {
             // Windows doesn't support disconnecting, so currently just dispose of the device
-            Trace.Message($"DisconnectDeviceNative from device with ID:  {device.Id.ToHexBleAddress()}");            
+            Trace.Message($"DisconnectDeviceNative from device with ID:  {device.Id.ToHexBleAddress()}");
             if (device.NativeDevice is BluetoothLEDevice nativeDevice)
             {
                 _deviceOperationRegistry.Remove(device.Id.ToString());
--- a/Source/Plugin.BLE/Windows/Adapter.cs
+++ b/Source/Plugin.BLE/Windows/Adapter.cs
@@ -97,10 +97,11 @@ namespace Plugin.BLE.Windows
             // var genericProfileGattSession = await GattSession.FromDeviceIdAsync(deviceId);            
             // bool success = genericProfileGattSession.MaintainConnection = true;
 
-            var servicesResult = await dev.NativeDevice.GetGattServicesAsync(BluetoothCacheMode.Uncached);
+            //var servicesResult = await dev.NativeDevice.GetGattServicesAsync(BluetoothCacheMode.Uncached);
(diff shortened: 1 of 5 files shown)


AskBojesen:
Hi @janusw , I found a bug in the Windows implementation I made recently.
Do you want be to create an issue also - or just this PR?
(It is not an issue in the released 3.0.0, since my Windows implementation came afterwards)

janusw:
> Do you want me to create an issue also - or just this PR?

PR only is fine with me in this case, but maybe you can describe the problem a bit better? You explained the situation in which it occurs, but not what exactly happens.

AskBojesen:
The problem wasbelow lines.... In Windows the BLE connection is established when asking for services.
Using the device.GetServicesAsync() which has a lot of unwanted sideeffects caused problems when reconnecting.
Eg. number of KnownServices got to high - setting BleImplementation.CacheModeGetServices - and then restoring was a hack I made to get something working - now I have a solution I understand, that also works better

`BluetoothCacheMode restoremode = BleImplementation.CacheModeGetServices;
BleImplementation.CacheModeGetServices = BluetoothCacheMode.Uncached;
var services = device.GetServicesAsync(cancellationToken).Result;
BleImplementation.CacheModeGetServices = restoremode;`

janusw (commented):
Hm, getting the services seems like a rather indirect way of triggering a connection. Isn't there a more direct method for connecting to a device on Windows?

On all other platforms, connecting to a device and getting the services are two separate steps ...

AskBojesen:
> Hm, getting the services seems like a rather indirect way of triggering a connection. Isn't there a more direct method for connecting to a device on Windows?
> 
> On all other platforms, connecting to a device and getting the services are two separate steps ...

Nope... And the previous implementation did the same...

In Wndows one get this object :

https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothledevice

You can see, that is no connect method...

The object is created by calling this method:

https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothledevice.frombluetoothaddressasync

We you can read:

_Creating a [BluetoothLEDevice](https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothledevice?view=winrt-22621) object by calling this method alone doesn't (necessarily) initiate a connection. To initiate a connection, set [GattSession.MaintainConnection](https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattsession.maintainconnection) to true, or call an uncached service discovery method on BluetoothLEDevice, or perform a read/write operation against the device._

AskBojesen:
Well in the hindsight I would like to try out the hindsight _[GattSession.MaintainConnection](https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattsession.maintainconnection) = true_ 

I think it will be more clean - what I do not like about using the _NativeDevice.GetGattServicesAsync_ with the connect side effect is that I/We do not store the services we get.

Let me try out the _GattSession.MaintainConnection_ flag before we merge this PR

AskBojesen:
@janusw : I tried out the different approach in Windows, but it gave various problems.
See the added comments in code
This PR is not changing the existing connect method - we could make an issue exploring this further...

I vote for merging this at it is now

AskBojesen:
I managed to get the [connection working - inspired by the Android implementation.
Introducing ConnectInternal and DisconnectInternal in the device

AskBojesen:
@janusw : If you review and approve this one today - then I will be extra gratefully - because it's my birthday today!

I guess I will go thorough the entire Windows implementation in the next 1,5 month - because we need this to work in our application at work. My point is - I am testing my own stuff with a skeptic mind - this must simply be robust and functional for our needs. We appreciate the plugin much so far - it is alive and functional. You and former contributors have made something that actually works. Details in the interface could be improved - but the essential parts focus on get things working.
We appreciate that.

janusw:
> @janusw : If you review and approve this one today - then I will be extra gratefully - because it's my birthday today!

Happy birthday! 🎂 

(I'm not gonna jump through this hoop though 😉 - but I'll try have a careful look in the next days).

> I guess I will go through the entire Windows implementation in the next 1,5 month - because we need this to work in our application at work. My point is - I am testing my own stuff with a skeptic mind - this must simply be robust and functional for our needs.

Sounds good. I hope others will do some testing of the Windows implementation as well.

> We appreciate the plugin much so far - it is alive and functional. You and former contributors have made something that actually works. Details in the interface could be improved - but the essential parts focus on get things working. We appreciate that.

Thanks for the positive feedback :)

Most of that praise actually goes to other/former contributors. I am mostly trying to keep this project alive after the previous maintainer left, but I only authored a small part of the code (and basically none of the Windows implementation).

> Details in the interface could be improved

There is always room for improvement 😄. And we're always welcoming PRs which make this improvement happen. 😉

janusw:
> @janusw: I managed to get the connection working - inspired by the Android implementation. Introducing ConnectInternal and DisconnectInternal in the device.
> 
> (I think this is more clean that calling get services without needing the services)
> 
> It all started by investigating the MTU size (or MaxPduSize as Windows calls it)
> 
> Now I believe we have the best solution for the Windows implementation

Yes, I agree that this is better and cleaner than before.


> sorry for the editing during this PR, but I guess that is what review is in fact for - you got me thinking more about the solution.

That is absolutely what a review is meant for: Either to point out a specific problem, or to get you thinking deeper and trigger further improvements this way. I'm glad my review was a good one in that sense :)

Source/Plugin.BLE/Windows/Adapter.cs
```diff
@@ -82,13 +82,9 @@ protected override async Task ConnectToDeviceNativeAsync(IDevice device, Connect
...
 
-            // Calling the GetGattServicesAsync on the BluetoothLEDevice with uncached property causes the device to connect
-            BluetoothCacheMode restoremode = BleImplementation.CacheModeGetServices;
-            BleImplementation.CacheModeGetServices = BluetoothCacheMode.Uncached;
-            var services = device.GetServicesAsync(cancellationToken).Result;
-            BleImplementation.CacheModeGetServices = restoremode;
+            bool success = await dev.ConnectInternal(connectParameters, cancellationToken);
 
-            if (!services.Any() || nativeDevice.ConnectionStatus != BluetoothConnectionStatus.Connected)
+            if (!success)
```
janusw, line 87: You're adding extra whitespace here ...

Source/Plugin.BLE/Windows/Adapter.cs
```diff
@@ -153,12 +149,11 @@ private void Device_ConnectionStatusChanged(BluetoothLEDevice nativeDevice, obje
         protected override void DisconnectDeviceNative(IDevice device)
         {
             // Windows doesn't support disconnecting, so currently just dispose of the device
-            Trace.Message($"DisconnectDeviceNative from device with ID:  {device.Id.ToHexBleAddress()}");
+            Trace.Message($"DisconnectDeviceNative from device with ID:  {device.Id.ToHexBleAddress()}");
```
janusw, line 152: ... and here. Not a big deal, you might say. But if people later edit this file with an editor that automatically removes this, their commits will touch these lines, which can hinder blaming and git archeology.

janusw: Sorry for being pedantic here 🙈

AskBojesen: Its's ok - I have formatted the documents now

AskBojesen:
@janusw : I think your changes to the master branch is causing build status problems here.

janusw:
> @janusw : I think your changes to the master branch is causing build status problems here.

Huh, strange, the CI builds were working on master, not sure why they fail here now (on the Android targets, which you did not even touch). Maybe due to changes in the GHA runner images, will check later.

janusw:
I just re-triggered the CI builds on master, in order to see if the still work, or if they fail now ...

janusw:
> I just re-triggered the CI builds on master, in order to see if they still work, or if they fail now ...

Indeed the Windows build fails with the same error now as on your branch, so this is obviously not your fault (and not your problem). I will look into it ...

janusw:
Fixed the CI on master and rebased your branch. I hope it should succeed now ...

janusw (approved):
The CI builds succeed now. This PR is ok with me.

janusw:
Merged. Thanks for the contribution!

PR #781, opened 2023-11-21
Scan Response Advertisement Records Implemented in Windows

AskBojesen:
We have a need to distinguish Advertisement Records  from a normal advertisement package and a scan response
This solves this need in Windows

### DIFF
--- a/Source/BLE.Client/BLE.Client.WinConsole/BLE.Client.WinConsole.csproj
+++ b/Source/BLE.Client/BLE.Client.WinConsole/BLE.Client.WinConsole.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net7.0-windows10.0.19041</TargetFramework>
+    <TargetFramework>net8.0-windows10.0.19041</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
   </PropertyGroup>
--- a/Source/Plugin.BLE/Shared/Contracts/IDevice.cs
+++ b/Source/Plugin.BLE/Shared/Contracts/IDevice.cs
@@ -48,6 +48,15 @@ namespace Plugin.BLE.Abstractions.Contracts
         IReadOnlyList<AdvertisementRecord> AdvertisementRecords { get; }
 
         /// <summary>
+        /// All the advertisment records received as scan responses
+        /// For example:
+        /// - Advertised Service UUIDS
+        /// - Manufacturer Specifc data
+        /// - ...
+        /// </summary>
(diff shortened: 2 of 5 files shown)


AskBojesen:
@janusw : This is my suggested implementation extending the interface.
A different approach would be to merge the two lists into one and add a flag to the AdvertisementRecord class where it noted `ScanResponse = true/false`
A maybe something else ?

janusw:
Some comments and questions ...

> @janusw : This is my suggested implementation extending the interface. A different approach would be to merge the two lists into one and add a flag to the AdvertisementRecord class where it noted `ScanResponse = true/false`

Actually I think using a boolean flag would be a better solution to this problem. I don't like the duplication of the `AdvertisementRecords` list. Many applications may not care at all if a given advertisement is a scan response or not, so they would need to look through both lists. A new flag is easier to ignore and more backwards compatible.

Second point: You're extending the `IDevice` interfacing, but only implement the new parts on Windows. Would it be possible to do the same on other platforms? I.e. could one populate a `ScanResponse` flag also on Android/iOS/MacOS?

And a last question: Why do you need to distinguish scan responses from 'other' advertisements? What's your use case?

AskBojesen:
Ok. My need is to get the contents of a Scan Response, knowing that it is a scan response
In the current windows implementation the Scan Response is replacing the other advertisement records.
We could extend the `AdvertisementRecord` class to include the PDU type of the packet it was inside

image

Ref:
https://www.bluetooth.com/bluetooth-resources/the-bluetooth-low-energy-primer/?utm_source=internal&utm_medium=blog&utm_campaign=technical&utm_content=the-bluetooth-low-energy-primer

Also here Ref: https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.advertisement.bluetoothleadvertisementtype

janusw:
> We could extend the `AdvertisementRecord` class to include the PDU type of the packet it was inside

Yes, that makes sense, I guess. I'd hope that one could also determine the PDU type on other platforms ...

AskBojesen:
> > We could extend the `AdvertisementRecord` class to include the PDU type of the packet it was inside
> 
> Yes, that makes sense, I guess. I'd hope that one could also determine the PDU type on other platforms ...

Yes - I totally agree :-) But if we agree on one implementation as a start - then we can hope that we can implement for Apple and Android also afterwards. We can also make it on attribute to the new field in the AdvertisementRecord - stating - only implemented in Windows - like it is written in a lot of places for Android only attributes availably in the Shared interface.

And again maybe  - will not even this myself in next week if we solves the issue otherwise.
I will mark this PR as draft... Since I am not ready

AskBojesen:
Ok - closing this - will open new if relevant.
Thanks for a nice collaboration

janusw:
> Ok - closing this - will open new if relevant.

:+1:

> Thanks for a nice collaboration

dito 😃

PR #784, opened 2023-11-29
Turn Bluetooth Radio On/Off

AskBojesen:
In the Windows Implementation I have added a `TrySetStateAsync`.
It works nicely in the Windows tool I have, but I guess security stuff doesn't allow this for Android + iOS, so I find no need to extend the interface,
but it is nice that the Windows implementation reuse the radio instance.

### DIFF
--- a/Source/Plugin.BLE/Shared/Contracts/IBluetoothLE.cs
+++ b/Source/Plugin.BLE/Shared/Contracts/IBluetoothLE.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
 using Plugin.BLE.Abstractions.EventArgs;
 
 namespace Plugin.BLE.Abstractions.Contracts
@@ -28,6 +29,13 @@ namespace Plugin.BLE.Abstractions.Contracts
         /// <c>true</c> if <see cref="State"/> is <c>BluetoothState.On</c>
         /// </summary>
         bool IsOn { get; }
+
+        /// <summary>
+        /// Try set the state of the Bluetooth on/off
+        /// </summary>
+        /// <param name="on"></param>
+        /// <returns>true if the the method executed with success otherwice false</returns>
+        Task<bool> TrySetStateAsync(bool on);
 
         /// <summary>
         /// Adapter to that provides access to the physical bluetooth adapter.
--- a/Source/Plugin.BLE/Shared/Contracts/IBluetoothLE.cs
+++ b/Source/Plugin.BLE/Shared/Contracts/IBluetoothLE.cs
@@ -32,6 +32,7 @@ namespace Plugin.BLE.Abstractions.Contracts
 
         /// <summary>
         /// Try set the state of the Bluetooth on/off
+        /// 2024-01-14: Only supported in Windows
         /// </summary>
         /// <param name="on"></param>
         /// <returns>true if the the method executed with success otherwice false</returns>
--- a/Source/Plugin.BLE/Android/BleImplementation.cs
+++ b/Source/Plugin.BLE/Android/BleImplementation.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
 using Android.App;
 using Android.Bluetooth;
 using Android.Content;
@@ -81,5 +82,11 @@ namespace Plugin.BLE
 
         protected override IAdapter CreateNativeAdapter()
             => new Adapter(_bluetoothManager);
+
+        public override Task<bool> TrySetStateAsync(bool on)
+        {
+            Abstractions.Trace.Message("WARNING TrySetStateAsync is not implemented for Android");
+            return Task<bool>.FromResult(false);
+        }
     }
 }
--- a/Source/Plugin.BLE/Apple/BleImplementation.cs
+++ b/Source/Plugin.BLE/Apple/BleImplementation.cs
@@ -4,6 +4,7 @@ using Plugin.BLE.Abstractions;
 using Plugin.BLE.Abstractions.Contracts;
 using Plugin.BLE.Extensions;
 using Plugin.BLE.iOS;
+using System.Threading.Tasks;
 
 namespace Plugin.BLE
 {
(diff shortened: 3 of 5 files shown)


axa88:
> but I guess security stuff doesn't allow this for Android 

I may not know what im talking about here, but is it no longer possible to use [ActionRequestEnable](https://learn.microsoft.com/en-us/dotnet/api/android.bluetooth.bluetoothadapter.actionrequestenable) to turn on/off bluetooth for `Android`?

Used this in` Xamarin.Android` and worked for me, but is implementing such platform specific code not available or in Maui?

AskBojesen:
> I may not know what im talking about here, but is it no longer possible to use [ActionRequestEnable](https://learn.microsoft.com/en-us/dotnet/api/android.bluetooth.bluetoothadapter.actionrequestenable) to turn on/off bluetooth for `Android`?
> 
> Used this in` Xamarin.Android` and worked for me, but is implementing such platform specific code not available for this library?

Yes! - I am the one who has limited knowledge about Android/iOS API. It seems we could extend the interface with a common method - what do you say @janusw - should we extend the interface for turning Bluetooth On/Off ?

AskBojesen:
Hallo @janusw , are you busy? I hope so...
When do you have time for this nice BLE package again?

axa88:
>  should we extend the interface for turning Bluetooth On/Off ?

@AskBojesen  curious why you wouldnt? Isnt the goal cross-platform parity?

janusw:
Sorry for the late reply here. Indeed I was pretty busy with other stuff.

> It seems we could extend the `IBluetoothLE` interface with a common method - what do you say @janusw - should we extend the interface for turning Bluetooth On/Off ?

Yes, if we implement toggling of the BT radio state, it should have a generic interface and should be implemented on all platforms where it's possible (so probably Windows and Android at least). Also the documentation should state on which platforms it's available.

For iOS it does not seem to be directly possible. For some discussion see e.g. https://stackoverflow.com/questions/47891941/open-bluetooth-settings-programmatically-from-uialertcontroller. Note that we are already using the `ShowPowerAlert` option when initializing the CBCentralManager, which opens the settings in case BT is not enabled. Possibly one could use this mechanism? However, it will only work for initially enabling BT (not disabling), so maybe one should better have separate `TryEnable` and `TryDisable` methods instead of a `TrySetState`?

AskBojesen:
@janusw : Thanks for your time! Should I extend the shared interface with this method in this branch?:

`public async Task<bool> TrySetStateAsync(bool on)`

@axa88  could you then implement for Android?

axa88:
> @axa88 could you then implement for Android?

🤔 I could certainly try...
Perhaps with some support after you insert the scaffolding. I'm new here

janusw:
> @janusw : Thanks for your time! Should I extend the shared interface with this method in this branch?:
> 
> `public async Task<bool> TrySetStateAsync(bool on)`

Yeah, that's probably the most straightforward way to go.

I'm not sure if one should try to attempt an iOS solution as well (which might benefit from a different interface). But since that's a bit tricky, let's just go ahead and stick to Windows and Android with the interface you are proposing. :+1:

AskBojesen:
I will extend the shared interface and finish this pull request.
Then eg. you @axa88 can make a fork and implement for Android afterwards.
(I had no experience with this open source GitHub framework either until a few months ago, but learned it is a nice framework - simply make your own fork based on master - most likely best after this PR is finished closed, thereafter make a Pull Request)

AskBojesen:
@janusw : This PR is ready now in my opinion.

AskBojesen:
@janusw : Ping

Source/Plugin.BLE/Shared/Contracts/IBluetoothLE.cs
```diff
@@ -29,6 +30,13 @@ public interface IBluetoothLE
         /// </summary>
         bool IsOn { get; }
 
+        /// <summary>
+        /// Try set the state of the Bluetooth on/off
+        /// </summary>
+        /// <param name="on"></param>
+        /// <returns>true if the the method executed with success otherwice false</returns>
+        Task<bool> TrySetStateAsync(bool on);
```
janusw, line 38: I think it would be a good idea to document here that this is only implemented on Windows (for the time being).

AskBojesen:
I don't see why the linuxBuild shopuld fail - @janusw : Can you start check again?

janusw:
> I don't see why the linuxBuild should fail - @janusw : Can you start check again?

I'm pretty sure it has nothing to do with your changes. I'll re-run it ...

janusw:
Re-running didn't help unfortunately, it still fails. Probably due to a change in the runner image, most likely this one: https://github.com/actions/runner-images/issues/8952

The master branch also fails now (when re-run). I'll look into it ...

AskBojesen:
Thanks for the effort

janusw:
> The master branch also fails now (when re-run). I'll look into it ...

I have fixed the problem on master now.

AskBojesen:
I closed by mistake - thought is was merged - reponed.
@janusw : It is ready for merge

janusw (approved):

janusw:
I have merged this to master. Thanks for your contribution, @AskBojesen!

@axa88 I hope you can contribute the Android implementation in an upcoming PR. I can support and give some feedback.

axa88:
> I have merged this to master. Thanks for your contribution, @AskBojesen!
> 
> @axa88 I hope you can contribute the Android implementation in an upcoming PR. I can support and give some feedback.

I took a look when you posted this message. In all honestly the implementation within the library seemed a bit over my head, but in the month ive been working with this lib ive learned a lot.
After ive finished my PoC implementation I will come back and take a look if its still needed.

But the larger concern is that [this API has just been deprecated](https://stackoverflow.com/a/73883387/5122089) 
Is it worth the trouble?  Or perhaps another method to guide the user to settings to turn it on themselves?

janusw:
I just opened issue #821 to track the Android implementation (and keep it from getting forgotten).

PR #791, opened 2023-12-27
GHA: add a Linux build

janusw:
This PR adds a Linux build to the GitHub Actions config and disables the iOS and Mac targets when building on a Linux host, so that the libs and the sample app can be built on Linux successfully (but with the Android target only).


janusw:
The GHA jobs all run well, so I'll merge this.

PR #801, opened 2024-01-22
Connection parameters for windows

AskBojesen:
We need this in order to implement firmware update on devices.

https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothlepreferredconnectionparameters

### DIFF
--- a/Source/Plugin.BLE/Plugin.BLE.csproj
+++ b/Source/Plugin.BLE/Plugin.BLE.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras/3.0.44">
 	<PropertyGroup>
 		<TargetFrameworks>netstandard2.0;net7.0-android33.0;net7.0-ios;net7.0-maccatalyst;net8.0-android34.0;net8.0-ios;net8.0-maccatalyst</TargetFrameworks>
-		<TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);MonoAndroid13.0;Xamarin.iOS10;Xamarin.Mac20;uap10.0.19041;net7.0-windows10.0.19041;net8.0-windows10.0.19041</TargetFrameworks>
+		<TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);MonoAndroid13.0;Xamarin.iOS10;Xamarin.Mac20;uap10.0.19041;net7.0-windows10.0.19041;net8.0-windows10.0.19041;net8.0-windows10.0.22621.0</TargetFrameworks>
 		<AssemblyName>Plugin.BLE</AssemblyName>
 		<RootNamespace>Plugin.BLE</RootNamespace>
 		<Version>3.1.0-beta.1</Version>
--- a/Source/Plugin.BLE/Plugin.BLE.csproj
+++ b/Source/Plugin.BLE/Plugin.BLE.csproj
@@ -2,7 +2,7 @@
 	<PropertyGroup>
 		<TargetFrameworks>netstandard2.0;net7.0-android33.0;net8.0-android34.0</TargetFrameworks>
 		<TargetFrameworks Condition="!$([MSBuild]::IsOSPlatform('linux'))">$(TargetFrameworks);net7.0-ios;net7.0-maccatalyst;net8.0-ios;net8.0-maccatalyst</TargetFrameworks>
-		<TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);MonoAndroid13.0;Xamarin.iOS10;Xamarin.Mac20;uap10.0.19041;net7.0-windows10.0.19041;net8.0-windows10.0.19041</TargetFrameworks>
+		<TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);MonoAndroid13.0;Xamarin.iOS10;Xamarin.Mac20;uap10.0.19041;net7.0-windows10.0.19041;net8.0-windows10.0.19041;net8.0-windows10.0.22621.0</TargetFrameworks>
 		<AssemblyName>Plugin.BLE</AssemblyName>
 		<RootNamespace>Plugin.BLE</RootNamespace>
 		<Version>3.1.0-beta.1</Version>
--- a/Source/Plugin.BLE/Plugin.BLE.csproj
+++ b/Source/Plugin.BLE/Plugin.BLE.csproj
@@ -1,7 +1,8 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras/3.0.44">
 	<PropertyGroup>
-		<TargetFrameworks>netstandard2.0;net7.0-android33.0;net7.0-ios;net7.0-maccatalyst;net8.0-android34.0;net8.0-ios;net8.0-maccatalyst</TargetFrameworks>
-		<TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);MonoAndroid13.0;Xamarin.iOS10;Xamarin.Mac20;uap10.0.19041;net7.0-windows10.0.19041;net8.0-windows10.0.19041;net8.0-windows10.0.22621.0</TargetFrameworks>
+		<TargetFrameworks>netstandard2.0;net7.0-android33.0;net8.0-android34.0</TargetFrameworks>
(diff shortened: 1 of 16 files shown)


AskBojesen:
Hi @janusw , this might seem like a big PR at first glance, but it is not.
Most changes are in the demo client - the change is straight forward to add connection parameters for Windows using the existing interface.

AskBojesen:
I need the `UpdateConnectionParameters` method in order to change the parameters for an already connected device - eg. before initiating firmwareupload

janusw:
> Hi @janusw , this might seem like a big PR at first glance, but it is not.

Yeah, the changes in the shared code are manageable. I'll try to have a closer look at the weekend ...

Source/Plugin.BLE/Plugin.BLE.csproj
```diff
@@ -2,7 +2,7 @@
 	<PropertyGroup>
 		<TargetFrameworks>netstandard2.0;net7.0-android33.0;net8.0-android34.0</TargetFrameworks>
 		<TargetFrameworks Condition="!$([MSBuild]::IsOSPlatform('linux'))">$(TargetFrameworks);net7.0-ios;net7.0-maccatalyst;net8.0-ios;net8.0-maccatalyst</TargetFrameworks>
-		<TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);MonoAndroid13.0;Xamarin.iOS10;Xamarin.Mac20;uap10.0.19041;net7.0-windows10.0.19041;net8.0-windows10.0.19041</TargetFrameworks>
+		<TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);MonoAndroid13.0;Xamarin.iOS10;Xamarin.Mac20;uap10.0.19041;net7.0-windows10.0.19041;net8.0-windows10.0.19041;net8.0-windows10.0.22000.0</TargetFrameworks>
```
janusw, line 5: Why do you add an additional Windows API version here, instead of updating the existing one? We anyway support Windows versions down to 10.0.17763.0, right? In there a reason we need two net8.0-windows targets?

Also, why not do the same for .NET7? Currently, the net7.0 target would not be able to use the newer API features. If .NET7 supports API 10.0.222, I would recommend to update it as well ...

Source/Plugin.BLE/Shared/ConnectParameter.cs
```diff
@@ -26,10 +32,14 @@ public struct ConnectParameters
         /// </summary>
         /// <param name="autoConnect">Android only: Whether to directly connect to the remote device (false) or to automatically connect as soon as the remote device becomes available (true). The default is false.</param>
         /// <param name="forceBleTransport">Android only: For Dual Mode device, force transport mode to LE. The default is false.</param>
-        public ConnectParameters(bool autoConnect = false, bool forceBleTransport = false)
+        public ConnectParameters(
+            bool autoConnect = false,
+            bool forceBleTransport = false,
+            ConnectionParameterSet connectionParameterSet = ConnectionParameterSet.None)
```
janusw, line 38: Maybe add a brief documentation of the third param, analogous to the existing two? There are warnings about this in GHA.

Source/Plugin.BLE/Shared/ConnectionParameterSet.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Plugin.BLE.Abstractions
+{
+    /// <summary>
+    /// Only supported in Windows. Mapped to this 
+    /// https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothlepreferredconnectionparameters
+    /// </summary>
+    public enum ConnectionParameterSet
+    {
+        /// <summary>
+        /// Not setting any preffered connection type
```
janusw, line 10: typo: preffered -> preferred

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
```diff
@@ -120,10 +120,22 @@ public interface IDevice : IDisposable
...
         bool SupportsIsConnectable { get; }
-        
+
         /// <summary>
         /// Gets the bonding state of a device.
         /// </summary>
         DeviceBondState BondState { get; }
+
+        /// <summary>
+        /// Updates the connection paramaters if already connected
```
janusw, line 130: paramaters -> parameters

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
```diff
@@ -120,10 +120,22 @@ public interface IDevice : IDisposable
...
+        /// </summary>
+        /// <remarks>
+        /// Only implemented for Windows
+        /// </remarks>
+        /// <param name="connectParameters">Connection parameters. Contains platform specific parameters needed to achieved connection. The default value is None.</param>
```
janusw, line 135: achieved -> achieve

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
```diff
@@ -120,10 +120,22 @@ public interface IDevice : IDisposable
...
+        /// <returns>
+        /// The Result property will contain a boolean that inticates if the update was successful.
```
janusw, line 137: inticates -> indicates

Source/Plugin.BLE/Shared/DeviceBase.cs
```diff
@@ -260,12 +260,19 @@ public override int GetHashCode()
...
         public abstract bool SupportsIsConnectable { get; }
-        
+
         /// <summary>
         /// Gets the <see cref="DeviceBondState"/> of the device.
         /// </summary>
         protected abstract DeviceBondState GetBondState();
 
+        /// <summary>
+        /// Updates the connection paramaters if already connected
```
janusw, line 270: paramaters -> parameters

janusw (commented):
Looks pretty good overall, thanks for the contribution. A few minor comments/questions below ...

Source/Plugin.BLE/Plugin.BLE.csproj
AskBojesen: I guess if we only have net8.0-windows10.0.22000.0 and not net8.0-windows10.0.19041, then if someone is on net8.0-windows10.0.19041, then they cannot use this package.  - and likewise for .Net7.0

The latest is in fact net8.0-windows10.0.22621.0, which I use

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
AskBojesen: Fixed - sorry for my spelling

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
AskBojesen: Fixed - sorry for my spelling

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
AskBojesen: Fixed - sorry for my spelling

Source/Plugin.BLE/Shared/ConnectionParameterSet.cs
AskBojesen: Fixed - sorry for my spelling

Source/Plugin.BLE/Plugin.BLE.csproj
AskBojesen: I resolved by mistake I guess - and reopened. We don't have a conclusion.
I agree that it is currently compiled to many platforms - at least I simplified the codebase so we only have one Windows implementation now. As a NuGet provider do you then have any usage statistics of the platforms?

janusw: > If we only have net8.0-windows10.0.22000.0 and not net8.0-windows10.0.19041, then if someone is on net8.0-windows10.0.19041, then they cannot use this package. - and likewise for .Net7.0

Ah, right. I missed that.


> You decide what is best for everybody - what I need is something greater or equal to net8.0-windows10.0.22000.0

I guess it's no big problem to have an additional target, but we have quite a number by now. At some point we may have to clean up some older ones.

janusw: > As a NuGet provider do you then have any usage statistics of the platforms?

Unfortunately not. nuget.org only reports the download numbers of the packet, not on which platforms it's used.

janusw (approved):
This is ok with me by now. Thanks for the documentation and spelling fixes (and sorry for being pedantic there, but I think also the documentation deserves a bit of love 😜 ).

AskBojesen:
Ï will have to enable spellchecker extension in the vs2022  - thanks for the review

PR #805, opened 2024-02-01
Added services for citysports treadmill

### LINKED ISSUE #804: Missing Characteristics on CitySports Treadmill
## Context

I'm trying to send requests to every discovered characteristic whilst doing a packet capture so that I can work out what UUIDs correspond with these handles:

image

## Demonstration Code

https://github.com/hughesjs/OpenEQiSports/tree/6bd2ec243acc91637b96031829c55f5e1087440b/proof-of-concept/EqiPoc

(This isn't a standalone repo, but the only thing in it so far is a PoC displaying this issue so it may as well be as long as you look at the linked commit specifically)

## Steps to reproduce

1. Launch app with debugging
2. Either break on TreadmillController:46 and look at the characteristics found (specifically for `service.Name == "Fitness Machine"`
3. Or press debug and look at the console output

## Expected behavior
According to nRF Connect, the "Fitness Machine" service has 14 characteristics.

image
image

## Actual behavior
Only 9 characteristics are discovered.

### Crashlog

N/A

## Configuration

**Version of the Plugin:** 3.0.0

**Platform:** iOS 17.3

**Device:** iPhone 15 Pro

hughesjs:
Fixes #804


Source/Plugin.BLE/Shared/KnownServices.cs
```diff
@@ -63,6 +63,7 @@ public static KnownService Lookup(Guid id)
             new KnownService("TI SensorTag Connection Control", Guid.ParseExact("f000ccc0-0451-4000-b000-000000000000", "d")),
             new KnownService("TI SensorTag OvertheAir Download", Guid.ParseExact("f000ffc0-0451-4000-b000-000000000000", "d")),
             new KnownService("TXRX_SERV_UUID RedBearLabs Biscuit Service", Guid.ParseExact("713d0000-503e-4c75-ba94-3148f18d941e", "d")),
+            new KnownService("CITYSPORTS Treadmill", Guid.ParseExact("00001826-0000-1000-8000-00805f9b34fb", "d")),
```
janusw, line 66: The BLE [Assigned Numbers](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Assigned_Numbers/out/en/Assigned_Numbers.pdf?v=1706903097464) document lists the ID 0x1826 as "Fitness Machine service" (which is also how nRF connect shows it). I think it would be better to use this official and generic name, instead of using a specific vendor name (which is only one of many who may implement it).

janusw (commented):
Two questions:
1. Your screenshot in #804 seems to show six characteristics that are recognized by nRF Connect (if I'm counting correctly). Why do you add only two of them, and not all six?
2. Why do you use different names than the ones shown by nRF Connect? (I think those are the "standard" ones)

Source/Plugin.BLE/Shared/KnownServices.cs
hughesjs: Honestly, because I have no idea where these GUIDs come from and the fact that you had a hardcoded list made me think they were just manufacturer generated rather than assigned... I'll tweak it. 

As for the rest, I'm not adding them because I've not found any use for them or been able to work out what they actually do, so I don't want to put my name against something I can't attest to

janusw: > Honestly, because I have no idea where these GUIDs come from and the fact that you had a hardcoded list made me think they were just manufacturer generated rather than assigned... I'll tweak it.

thx

> As for the rest, I'm not adding them because I've not found any use for them or been able to work out what they actually do, so I don't want to put my name against something I can't attest to

Very reasonable attitude 😃, but at least for those that are included in the BLE Assigned Numbers document (and listed as "known" characteristics by nRF connect) there is no need to worry: They are standard characteristics with a well-defined name. If you are unsure what they do or how they are supposed to be used, maybe the Fitness Machine service specs will help to clarify that: https://www.bluetooth.com/specifications/specs/fitness-machine-service-1-0/

For those listed as "Unknown Characteristic" by nRF Connect, it's certainly a good idea to be a bit more careful: They are probably non-standard / vendor-specific characteristics, so maybe let's rather ignore them for now (unless you find some documentation that clearly defines their function).

hughesjs: Well, that's a handy document I wish I'd known about last week before I reversed the entire protocol with pcaps... I'll add the rest if I get a chance, not sure when that'll be, maybe merge what we have here for now?

janusw: Nevermind, I'll just finish just up myself. No big deal ...

janusw:
While we're at it, I'm also cleaning up and extending the KnownServices a little more ...

Source/Plugin.BLE/Shared/KnownServices.cs
janusw: I'll simply merge this now. Let me know if you find any problem in hindsight.

janusw (approved):

hughesjs:
Thanks for finishing that off for me

PR #806, opened 2024-02-04
GHA: build Maui client on Windows

janusw:
I recently noticed that a couple of our sample clients are not yet inlcuded in the cake build script and therefore not built on Windows via GHA. Some of them I have already added on master, missing at this point is only BLE.Client.Maui, which is added by the PR. However, the Windows build currently fails (on the net8.0-windows target) with this error, which I don't quite understand:

```
Unable to optimize assemblies for performance: a valid runtime package was not found.
Either set the PublishReadyToRun property to false, or use a supported runtime identifier when publishing.
When targeting .NET 6 or higher, make sure to restore packages with the PublishReadyToRun property set to true.
```

I'm not even sure why anything is supposed to be 'published' here, because we actually just want to build the sample client, nothing more.

@AskBojesen Since you are our Windows expert: Do you understand this error and do you have an idea what's the best way to get rid of it?

### DIFF
--- a/.build/build.cake
+++ b/.build/build.cake
@@ -76,6 +76,7 @@ Task("BuildClients")
   BuildProject("BLE.Client", "BLE.Client.UWP", Path.Combine("clients", "uwp"));
   // .NET 7/8
   BuildProject("BLE.Client", "BLE.Client.WinConsole", Path.Combine("clients", "wincon"));
+  BuildProject("BLE.Client", "BLE.Client.Maui", Path.Combine("clients", "maui"));
 });
 
 Task("Clean").Does (() =>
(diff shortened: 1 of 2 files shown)


janusw:
> ```
> Unable to optimize assemblies for performance: a valid runtime package was not found.
> Either set the PublishReadyToRun property to false, or use a supported runtime identifier when publishing.
> When targeting .NET 6 or higher, make sure to restore packages with the PublishReadyToRun property set to true.
> ```

Following the first suggestion of this error message, namely setting `PublishReadyToRun` to false, actually seems to fix the error, but triggers a different one:

```
error : Packaged .NET applications with an app host exe cannot be ProcessorArchitecture neutral.
Please specify a RuntimeIdentifier or a Platform other than AnyCPU.
```

I'm already using `win-x64` as a RuntimeIdentifier for the Windows case, so I'm a bit lost here.

@AskBojesen Do you have any advice? What am I missing?

janusw:
Alright, looks like I figured it out after some trial-and-error. That this kind of configuration magic is necessary looks like some sort of quirk in the .NET build system to me (certainly not the only one) ...

PR #809, opened 2024-02-07
Fix ConnectionLost -> Connect for Windows

AskBojesen:
Reproduce the error:
1. Connect
2. ConnectLost
3. Connect

Thereafter ConnectionChanged event was fired twice subsequently

@janusw : Yet another small PR for the Windows stability

### DIFF
--- a/Source/Plugin.BLE/Windows/Adapter.cs
+++ b/Source/Plugin.BLE/Windows/Adapter.cs
@@ -19,10 +19,10 @@ namespace Plugin.BLE.Windows
         private BluetoothLEAdvertisementWatcher _bleWatcher;
 
         /// <summary>
-        /// Registry used to store device instances for pending operations : disconnect
+        /// Registry used to store device instances for pending disconnect operations
         /// Helps to detect connection lost events.
         /// </summary>
-        private readonly IDictionary<string, IDevice> _deviceOperationRegistry = new ConcurrentDictionary<string, IDevice>();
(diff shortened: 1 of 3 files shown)


Source/Plugin.BLE/Windows/Adapter.cs
```diff
@@ -144,11 +144,8 @@ private void Device_ConnectionStatusChanged(BluetoothLEDevice nativeDevice, obje
                 }
 
                 // fire the correct event (DeviceDisconnected or DeviceConnectionLost)
-                HandleDisconnectedDevice(isNormalDisconnect, disconnectedDevice);
-                if (isNormalDisconnect)
-                {
-                    nativeDevice.ConnectionStatusChanged -= Device_ConnectionStatusChanged;
-                }
+                HandleDisconnectedDevice(isNormalDisconnect, disconnectedDevice);
```
janusw, line 147: There is no real change in line 147. Apparently you're adding unnecessary whitespace at the end.

AskBojesen:
I ended up with a quite big change in the adapter.cs , but I am really confident this now solves my issues since I tested a lot in the two cases:

1. Connect -> Disconnect ->Connect -> Disconnect ->Connect -> Disconnect 

2. Connect -> LostConnection ->Connect -> LostConnection -> Connect -> LostConnection

AskBojesen:
@janusw: Can you look at this PR?

janusw (approved):
I'm not able to judge if this does the right thing in every circumstance, but I don't see anything obviously wrong with your code either.
So if you tested it thoroughly, the PR is ok with me.

PR #818, opened 2024-02-27
Optional Autopair when connecting in Windows

AskBojesen:
This feature is to solve the following scenario for producing our BLE devices using a Windows application as production test tool:

1. N devices (typical N=5) set in pairing mode
2. A Windows application pair and connects in one go without user popup interaction
3. The Windows application can now read the GATT table and services while being paired and connected

**This PR solves this scenario**

**The problem with the existing implementation** 
If I first use the `DeviceInformationCustomPairing.PairAsync` API ref https://learn.microsoft.com/en-us/uwp/api/windows.devices.enumeration.deviceinformationcustompairing.pairasync?view=winrt-22621 - then I can pair - but an instance somewhere in the Windows api is then connected.
I then have to turn bluetooth off/on or close my application instance in order to disconnect so I can connect again with the PlugIn.BLE interface

@janusw : When this is done - then I am feature complete with our needs for the Windows implementation for now...
(I will fix bugs of course...)


axa88:
@AskBojesen

Sorry, dont mean to hijack your commit, but Im out of the loop.
Ive begun eval this lib on windows recently, and reading your commit notes it seems this expands on pairing/bonding in windows, but i wasn't aware paring/bonding was implemented in the first place, is there a version where it is?

AskBojesen:
> @AskBojesen
> 
> Sorry, dont mean to hijack your commit, but Im out of the loop. Ive begun eval this lib on windows recently, and reading your commit notes it seems this expands on pairing/bonding in windows, but i wasn't aware paring/bonding was implemented in the first place, is there a version where it is?

The existing paring/bonding methods for the IAdapter are Not implemented for Windows.
You can see that here: https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/master/Source/Plugin.BLE/Windows/Adapter.cs

I have no need for these two methods myself in our project - but I can easily implement.
I will do that in another PR - just for you :-)

```
protected override IReadOnlyList<IDevice> GetBondedDevices()
        {
            return null; // not supported
        }

 public override Task BondAsync(IDevice device)
 {
     throw new NotImplementedException();
 }
```

axa88:
@AskBojesen lol, you and I have a bit different idea of the meaning of  _implement_ but you're technically accurate 🫡

axa88:
@AskBojesen 

Have you any knowledge regarding [this report](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/discussions/819)

AskBojesen:
@janusw : do you have time to review this PR?

janusw:
> **The problem with the existing implementation**
> If I first use the `DeviceInformationCustomPairing.PairAsync` API ref https://learn.microsoft.com/en-us/uwp/api/windows.devices.enumeration.deviceinformationcustompairing.pairasync?view=winrt-22621 - then I can pair - but an instance somewhere in the Windows api is then connected. I then have to turn bluetooth off/on or close my application instance in order to disconnect so I can connect again with the PlugIn.BLE interface

Just to make sure I understand correctly what you mean by "existing implementation": Is this a problem with the existing implementation *before* this PR, which is solved by the PR? Or is it a problem which still affects the existing implementation in this PR?

janusw:
Concerning the logical ordering, this PR seems more like an extension of #820 to me: The latter implements the basic interfaces for pairing/bonding, and this one adds the possibility for automatic pairing upon connection, right?
Therefore I'd like to have #820 merged first, before I approve this one.

Source/Plugin.BLE/Windows/Device.cs
```diff
@@ -168,9 +165,25 @@ public async Task<bool> ConnectInternal(ConnectParameters connectParameters, Can
...
+                    {
+                        Trace.Message($"Pairing with {deviceInformation.Name} ({deviceInformation.Id})");
+                        deviceInformation.Pairing.Custom.PairingRequested += Custom_PairingRequested;
+                        DevicePairingResult result = await deviceInformation.Pairing.Custom.PairAsync(
+                            DevicePairingKinds.ConfirmOnly, DevicePairingProtectionLevel.Encryption);
+                        Trace.Message($"Pairing with {deviceInformation.Name} ({deviceInformation.Id}) result: {result.Status}");
+                        if (result.Status != DevicePairingResultStatus.Paired)
+                        {
+                            throw new Exception($"Could not pair: {result.Status}");
+                        }
```
janusw, line 183: This part looks pretty similar to the implementation of `BondAsync` in #820, but with some small differences (e.g. the parameters of `PairAsync`). Are there supposed to be any differences? If not, one could put this into a common method that is used in both places (for consistency, and to avoid code duplication).
As mentioned, I think #820 should be merged first, so that this PR can be added on top of it.

AskBojesen:
Handled in https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/pull/839

Issue #825, opened 2024-03-08
App .NET MAUI Crash on iOS


acaliaro:
To help us fix your issue, please provide the information in the below template. If something causes a crash, provide as much information as you can gather.
Just imagine: we do not know what you are doing! 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.

I am debugging a MAUI app on a real iOS device. Frequently the app crashes with the below stack trace. I can't say the precise instruction on which it crashes

## Steps to reproduce

1.

2.

3.

## Expected behavior
It should not crash

## Actual behavior
It crashes

### Crashlog

If something causes an exception paste full stack trace + Exception here

[0:] An error occurred: 'Object reference not set to an instance of an object.'. Callstack: '   at Plugin.BLE.iOS.Characteristic.NSErrorToGattStatus(NSError error)
   at Plugin.BLE.iOS.Characteristic.<>c__DisplayClass17_1.<StartUpdatesNativeAsync>b__7(Object sender, CBCharacteristicEventArgs args)
   at CoreBluetooth.CBPeripheral._CBPeripheralDelegate.UpdatedNotificationState(CBPeripheral peripheral, CBCharacteristic characteristic, NSError error) in /Users/builder/azdo/_work/1/s/xamarin-macios/src/build/dotnet/ios/generated-sources/CoreBluetooth/CBPeripheral.g.cs:line 528
   at UIKit.UIApplication.UIApplicationMain(Int32 argc, String[] argv, IntPtr principalClassName, IntPtr delegateClassName) in /Users/builder/azdo/_work/1/s/xamarin-macios/src/UIKit/UIApplication.cs:line 58
   at UIKit.UIApplication.Main(String[] args, Type principalClass, Type delegateClass) in /Users/builder/azdo/_work/1/s/xamarin-macios/src/UIKit/UIApplication.cs:line 94
   at daze.dazeapp.Program.Main(String[] args) in C:\Daze\daze.app.maui\daze.dazeapp\daze.dazeapp\Platforms\iOS\Program.cs:line 13
   at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
   at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)'

## Configuration

[...]

PR #827, opened 2024-03-17
Fix a NullReferenceException in Characteristic.NSErrorToGattStatus

janusw:
This is supposed to fix a small bug on iOS, see #825.

### DIFF
--- a/Source/Plugin.BLE/Apple/Characteristic.cs
+++ b/Source/Plugin.BLE/Apple/Characteristic.cs
@@ -111,7 +111,7 @@ namespace Plugin.BLE.iOS
 #endif
                         {
                             Trace.Message($"Read characterteristic value: {Value?.ToHexString()}");
-                            int resultCode = (args.Error == null) ? 0 : NSErrorToGattStatus(args.Error);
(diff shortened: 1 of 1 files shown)


janusw:
@acaliaro Can you confirm that the crashes are fixed by this?

janusw:
> @acaliaro Can you confirm that the crashes are fixed by this?

Note: You can use the nupkg built by GHA to check this ...

https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/actions/runs/8315448842?pr=827#artifacts

acaliaro:
> > @acaliaro Can you confirm that the crashes are fixed by this?
> 
> Note: You can use the nupkg built by GHA to check this ...
> 
> https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/actions/runs/8315448842?pr=827#artifacts

Hi @janusw It seems like your solution solves the problem. I did various tests and the exception never showed up again.

janusw:
> Hi @janusw It seems like your solution solves the problem. I did various tests and the exception never showed up again.

Awesome, thanks for checking 👍

PR #837, opened 2024-04-09
Maui client update

jeetee:
After testing the MAUI demo application on my mobile device (Sony XQ-DQ54 - Android14) and on my Windows11 PC it seemed to fall short in several areas:

* Doesn't fit on screen.
* Log Messages non-functional
* Default App Icon

This PR addressed those initial points and tidied up some obsolete code from the view models.

A subsequent branch/PR will look into adding Connecting to a device and reading out the services/characteristics from it.

Tested with .Net-8 and Android33/34 and Windows 10 & 11 only.

### DIFF
--- a/Source/BLE.Client/BLE.Client.Maui/Platforms/Android/AndroidManifest.xml
+++ b/Source/BLE.Client/BLE.Client.Maui/Platforms/Android/AndroidManifest.xml
@@ -1,9 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.companyname.ble.client.maui" android:versionCode="1" android:versionName="1.0">
-	<application android:allowBackup="true" android:supportsRtl="true" android:label="BLE.Client.Maui"></application>
+	<application android:allowBackup="true" android:supportsRtl="true" android:label="BLE.Client.Maui"
+				 android:icon="@mipmap/appicon" android:roundIcon="@mipmap/appicon_round"
+				 ></application>
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 	<uses-permission android:name="android.permission.INTERNET" />
-	<uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
+	<!--<uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
 	<uses-permission android:name="android.permission.BLUETOOTH" />
 	<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
 	<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
@@ -14,5 +16,8 @@
 	<uses-permission android:name="android.permission.BLUETOOTH_PRIVILEGED" />
 	<uses-permission android:name="android.permission.BLUETOOTH_SCAN" android:usesPermissionFlags="neverForLocation" />
 	<uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
-	<uses-permission android:name="android.permission.CONTROL_LOCATION_UPDATES" />
+	<uses-permission android:name="android.permission.CONTROL_LOCATION_UPDATES" />-->
+	<uses-feature android:name="android.hardware.bluetooth_le" android:required="true"/>
+	<uses-permission android:name="android.permission.BLUETOOTH_SCAN" android:usesPermissionFlags="neverForLocation" />
+	<uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
 </manifest>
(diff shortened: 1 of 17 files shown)


janusw:
Thanks for the contribution. And yes, you're right, the MAUI sample client is indeed rather basic right now and can some care. I'll have a look at this soon ...

jeetee:
Thanks for verifying. This is still early MAUI days for myself, but I believe the code can serve at least as an initial starting point for some.

I'm continuing this demo (even if just for my own purposes) towards reading the services/characteristics and then allow reading/subscribing/writing values. Once something stable comes out of that in the coming week(s) I'll open an additional PR. Progress can be followed here for those interested: https://github.com/jeetee/dotnet-bluetooth-le/tree/maui-client-device-browser

janusw:
I just tried your branch on Android and iOS and indeed the MAUI sample app looks a lot nicer and is much more usable on small screens now. Also the new log view is a useful addition. :+1:

For some reason I failed to get the sample running on MacOS. I'm pretty sure that was working at some point. It also seems to affect the master branch AFAICS, so it's probably not your fault. I will try to have another look at this problem soon.

Also I want to skim over the code a bit, but I think this PR is pretty close to approval already ...

Source/BLE.Client/BLE.Client.Maui/Platforms/Android/AndroidManifest.xml
```diff
@@ -14,5 +16,8 @@
 	<uses-permission android:name="android.permission.BLUETOOTH_PRIVILEGED" />
 	<uses-permission android:name="android.permission.BLUETOOTH_SCAN" android:usesPermissionFlags="neverForLocation" />
 	<uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
-	<uses-permission android:name="android.permission.CONTROL_LOCATION_UPDATES" />
+	<uses-permission android:name="android.permission.CONTROL_LOCATION_UPDATES" />-->
+	<uses-feature android:name="android.hardware.bluetooth_le" android:required="true"/>
+	<uses-permission android:name="android.permission.BLUETOOTH_SCAN" android:usesPermissionFlags="neverForLocation" />
+	<uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
```
janusw, line 22: I'm afraid you're breaking the permissions on older Android versions here: At least `BLUETOOTH` and `ACCESS_FINE_LOCATION` are needed on Android 11 and below. The other ones are probably ok to be removed.

janusw (changes requested):

Source/BLE.Client/BLE.Client.Maui/Platforms/Android/AndroidManifest.xml
jeetee: True.. I indeed started from my own use case scenario to test whether those "minimal" rights would cover my case.
I'll reinstate the other related permissions by checking the Android docs on them.

Source/BLE.Client/BLE.Client.Maui/Platforms/Android/AndroidManifest.xml
```diff
@@ -1,18 +1,23 @@
...
+		 for more information about which permissions to use for your target platform
+	-->
+	<uses-feature android:name="android.hardware.bluetooth_le" android:required="true" />
 	<uses-permission android:name="android.permission.BLUETOOTH_SCAN" android:usesPermissionFlags="neverForLocation" />
-	<uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
-	<uses-permission android:name="android.permission.CONTROL_LOCATION_UPDATES" />
+	<uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
+	<!-- Needed for Android 11 and lower -->
+	<uses-permission android:name="android.permission.BLUETOOTH" android:maxSdkVersion="30" />
+	<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" android:maxSdkVersion="30" />
```
janusw, line 17: Alright, this looks much better now. I verified that the app runs on an Android 11 device again. It properly asks for the location permission there.

Source/BLE.Client/BLE.Client.Maui/Platforms/Android/AndroidManifest.xml
```diff
@@ -1,18 +1,23 @@
...
+	<!-- Needed for some devices to be able to use LE local discovery on Android 10/11 -->
+	<uses-permission android:name="android.permission.BLUETOOTH_ADMIN" android:maxSdkVersion="30" />
+	<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" android:maxSdkVersion="30" />
+	<uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION" android:maxSdkVersion="30" />
+	<uses-permission android:name="android.permission.ACCESS_MOCK_LOCATION" android:maxSdkVersion="30" />
```
janusw, line 22: Strictly speaking, I don't think those would be necessary, but they certainly don't hurt either (most of them have already been there previously). So ok with me.

janusw (approved):

janusw:
Merged. Thanks for contributing!

PR #839, opened 2024-04-22
Windows: Handle that a device is already connected

AskBojesen:
This change allows the connect sequence to be established even though the device is already connected.
An already connection could be due to pairing, which connects the device.

@janusw: This ended up with a simple solution :-)
If merging this new simple PR, we can discard the old one, which I have now closed... (https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/pull/818)

### DIFF
--- a/Source/BLE.Client/BLE.Client.WinConsole/PluginDemos.cs
+++ b/Source/BLE.Client/BLE.Client.WinConsole/PluginDemos.cs
@@ -246,16 +246,19 @@ namespace BLE.Client.WinConsole
             DeviceInformation? deviceInformation = null;
             using (BluetoothLEDevice nativeDevice = await BluetoothLEDevice.FromBluetoothAddressAsync(bleAddressulong))
             {
-                nativeDevice.RequestPreferredConnectionParameters(BluetoothLEPreferredConnectionParameters.ThroughputOptimized);
-                nativeDevice.ConnectionStatusChanged += NativeDevice_ConnectionStatusChanged;
                 deviceInformation = await DeviceInformation.CreateFromIdAsync(nativeDevice.DeviceId);
             }
-            deviceInformation.Pairing.Custom.PairingRequested += Custom_PairingRequested;
-            Write("Pairing");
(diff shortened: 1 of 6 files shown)


AskBojesen:
Hi @janusw : If you merge this little one - then I am feature complete for our project - and I will stop poking you ;-)

Of course I will fix bugs and implement features in the Windows implementation if needed...

janusw (commented):
Indeed this looks a bit better to me than #818, also because it is much simpler. Your usage of `DeviceState.Limited` seems to match well how it's used on Android.

My only suggestion: Maybe you can update the documentation of this state a bit, mention Windows there in addition to Android, and improve the wording (currently it sounds a bit strange).

AskBojesen:
> Indeed this looks a bit better to me than #818, also because it is much simpler. Your usage of `DeviceState.Limited` seems to match well how it's used on Android.
> 
> My only suggestion: Maybe you can update the documentation of this state a bit, mention Windows there in addition to Android, and improve the wording (currently it sounds a bit strange).

What about now?

janusw:
> What about now?

Yes, sounds good. I guess one could even have a common description for Android and Windows, but let's not be too picky here :)

janusw (approved):
Ok with me. Thanks for the contribution!

AskBojesen:
> Ok with me. Thanks for the contribution!

@janusw : When do you plan to release 3.1.0 ?

janusw:
> @janusw : When do you plan to release 3.1.0 ?

My plan would be to tag an rc today and then release 3.1.0 within a week, if there are no showstoppers.

Does that sound reasonable to you, or do you need a stable release more urgently?

AskBojesen:
> > @janusw : When do you plan to release 3.1.0 ?
> 
> My plan would be to tag an rc today and then release 3.1.0 within a week, if there are no showstoppers.
> 
> Does that sound reasonable to you, or do you need a stable release more urgently?

It sounds really nice - we are not in a hurry - thanks!

axa88:
@AskBojesen 
Think this change would address this issue: #841
In fact, have you seen this problem at all. Im shocked it went past me so long undetected...

Issue #815, opened 2024-02-21
WinUI AdvertisementRecords missing advertised device records


axa88:
It appears to me there is an issue with collection of `IDevice.AdvertisementRecords` while scanning on `WinUI`.

I have found when a device advertises, `WinUI` the `AdvertisementRecords` will tend to intermittently lose/be overwritten by other `AdvertisementRecords` upon each subsequent `Adapter.DeviceAdvertised` events received.  And rarely if ever do all advertised data appear in the `AdvertisementRecords` at the same time.

Im unsure if this is a OS/hardware limitation but here are my observations.

Here is my esp32 device advertising 4 separate `AdvertisementRecord`s
![image](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/assets/13371986/5ab7cbd7-bcc4-4067-a476-660eb0b6ac36)

And Here is my MAUI app running on `Android` will list all 4 devices as they are found in `IDevice.AdvertisementRecords` upon `Adapter.DeviceDiscovered` event, as well as every subsequent `Adapter.DeviceAdvertised` event raised.

![image](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/assets/13371986/6c715cc1-bef4-4efb-9fd2-2dbb65b249f1)

While the same application running on `WinUI` will only list part of the advertised Records upon `Adapter.DeviceDiscovered` and each subsequent `Adapter.DeviceAdvertised` event raised.  Or may have a record missing completely. 
Here, the 4 Records were consistently split into 2 different distributions, 3 records, and 1 record:

![image](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/assets/13371986/daded88a-c348-44e2-8b49-4c2dedd63642)

I couldn't observe any pattern to the inconsistency, any event might return either the 1 record or 3 records.

Note the advertisement pack over 31 bytes; could this be getting broken up by the underlying BLE stack causing inconsistency on the different platforms?
So I wonder if this issue relates to #814 where Ive found `ScanFilterOptions` to also not work on `WinUI`

## Expected behavior
`IDevice.AdvertisementRecords` should consistently list all records provided by a device after initial `Adapter.DeviceDiscovered` and each subsequent `Adapter.DeviceAdvertised` event raised.

## Actual behavior
[...]

PR #845, opened 2024-05-05
Implement Adapter.SupportsExtendedAdvertising on Windows

janusw:
see the discussion in #815

### DIFF
--- a/Source/Plugin.BLE/Windows/Adapter.cs
+++ b/Source/Plugin.BLE/Windows/Adapter.cs
@@ -16,6 +16,7 @@ namespace Plugin.BLE.Windows
 {
     public class Adapter : AdapterBase
     {
+        private readonly BluetoothAdapter _bluetoothAdapter;
         private BluetoothLEAdvertisementWatcher _bleWatcher;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


axa88:
> @axa88 Could you check if this works well for you? (Testing on Windows is not fully impossible but rather effortful for me). There is a nupkg built by GHA which you can use for testing.

Im not entirely sure of what i need to confirm proper functionality.
But i do know i do not have Windows hardware that supports extended advertising...

janusw:
> > @axa88 Could you check if this works well for you? (Testing on Windows is not fully impossible but rather effortful for me). There is a nupkg built by GHA which you can use for testing.
> 
> Im not entirely sure of what i need to confirm proper functionality. But i do know i do not have Windows hardware that supports extended advertising...

Well, in that case it would be enough if you could simply confirm that `SupportsExtendedAdvertising` still returns `false`, as before (and there are no exceptions or errors).

I'll check if I have some Windows hardware that supports extended advertising.

axa88:
> Well, in that case it would be enough if you could simply confirm that `SupportsExtendedAdvertising` still returns `false`, as before (and there are no exceptions or errors).

Things continue to work as they did..

janusw:
> > Well, in that case it would be enough if you could simply confirm that `SupportsExtendedAdvertising` still returns `false`, as before (and there are no exceptions or errors).
> 
> Things continue to work as they did..

Thanks for the confirmation. I also managed to test this by now, but my Windows test device also returns `false` (an oldish Dell XPS 13 notebook). In any case this is enough to verify that the native API call works correctly, so I will merge this simple PR.

Issue #826, opened 2024-03-08
Windows: More than one lost connection does not work correctly


CodeMusher:
To help us fix your issue, please provide the information in the below template. If something causes a crash, provide as much information as you can gather.
Just imagine: we do not know what you are doing! 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.

## Steps to reproduce

1. I run the WinConsole application. Press D1 to listen for devices, Select my Device and press D9 to run Connect -> Loop: ConnectionLost -> Connect 
2. The first time I power cycle my device it clearly state : DisconnectedPeripheral by lost signal:

3. Power up the device => Normal behavior

4. Power cycle the device again. Now it states: DisconnectedPeripheral by user: 

## Expected behavior
It should not matter now many times I power cycle the device. It should always be DisconnectedPeripheral by lost signal

## Actual behavior
It works one time! The first power cycle operation causes the device to be put in the disconnectingRegistry

### Crashlog

If something causes an exception paste full stack trace + Exception here

## Configuration

**Version of the Plugin:**
3..0-beta-2.

**Platform:** e.g. iOS 10.1 / Android 4.4 / ... (including version!!! e.g. Android 5.1 / i0S 10)
Windows
**Device:** e.g. HTC Sensation /i Phone 7 ...
Laptop

smsissuechecker:
Hi @CodeMusher,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

janusw:
> **Version of the Plugin:** 3..0-beta-2.

This is supposed to be 3.1.0-beta.2, I assume?

CodeMusher:
Sorry, my bad! 3.1.0-beta.2 is correct

CodeMusher:
This could be solved by adding this line
disconnectingRegistry.Remove(id);
in the Device_ConnectionStatusChanged method of the Adapter class

janusw:
> This could be solved by adding this line disconnectingRegistry.Remove(id); in the Device_ConnectionStatusChanged method of the Adapter class

Hm, looks like that line is there already ...

https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/04a5d273324941ebc9069b179939e4b9b43dad40/Source/Plugin.BLE/Windows/Adapter.cs#L148-L151

[...]

Issue #788, opened 2023-12-21
BLE.Client.WinConsole demo app hangs on GetServicesAsync call


melihercan:
To help us fix your issue, please provide the information in the below template. If something causes a crash, provide as much information as you can gather.
Just imagine: we do not know what you are doing! 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.

## Steps to reproduce

1. Get main branch

2. Customize the program.cs so that it scans passively 3 times and then calls ShowNumberOfServices on [WGX-iBeacon](https://www.lucadentella.it/en/2018/03/19/wgx-ibeacons/) device:
```
using BLE.Client.WinConsole;
using Plugin.BLE;
using Plugin.BLE.Abstractions.Contracts;
using System;

Console.WriteLine("Hello, BLE World!");
using (var ct = new ConsoleTracer())
{
    ////Plugin.BLE.Abstractions.Trace.TraceImplementation = ct.GetPrefixedTrace("Plugin.BLE");
    var demo = new BleDemo(ct.GetPrefixedTrace("      DEMO"));

    for (int i=0; i<3; i++)
        await demo.DoTheScanning(ScanMode.Passive, 5000);
    await demo.ShowNumberOfServices("EB408B203E81");    // iBeacon
}
```
Also change the DoTheScanning debug messages a bit:
```
            ////Write("Bluetooth is on");
            Write("Scanning now for " + time_ms + " ms..." + " with mode:" + scanMode);            
```

## Expected behavior
1. It should scan and report the found BLE devices
2. Connect to WGX-iBeacon device
3. Get services and display them
4. Repeat 2 and 3

## Actual behavior
1. Scan is OK
2. Connection to the device takes a long time, ~30sec (please see the log below)!!!
3. Calls dev.GetServicesAsync and it hangs!!!

I have tried with different BLE devices and the issue is reproducible

### Crashlog
No crash just hanging. Here is the log:
```
Hello, BLE World!
16:31:40.687       DEMO - Scanning now for 5000 ms... with mode:Passive
16:31:41.226       DEMO - DeviceDiscovered: 5C4076784064 with Name = Bluetooth 5c:40:76:78:40:64
16:31:41.276       DEMO - DeviceDiscovered: E0189F0A7401 with Name = SimCPR sensor
[...]

PR #849, opened 2024-05-08
Windows: Connect -> ConnectionLost -> Reconnect

AskBojesen:
To fix: https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/issues/826

[demolog.txt](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/files/15251792/demolog.txt)

### DIFF
--- a/Source/BLE.Client/BLE.Client.WinConsole/PluginDemos.cs
+++ b/Source/BLE.Client/BLE.Client.WinConsole/PluginDemos.cs
@@ -23,6 +23,9 @@ namespace BLE.Client.WinConsole
         private readonly Action<string, object[]>? writer;
         private readonly List<IDevice> discoveredDevices;
         private bool scanningDone = false;
+        private ConsoleKey consoleKey = ConsoleKey.None;
+        private IDevice? reconnectDevice;
+        private CancellationTokenSource escKeyCancellationTokenSource = new CancellationTokenSource();
 
         public PluginDemos(Action<string, object[]>? writer = null)
         {
@@ -39,11 +42,6 @@ namespace BLE.Client.WinConsole
         private void Adapter_DeviceConnectionError(object? sender, Plugin.BLE.Abstractions.EventArgs.DeviceErrorEventArgs e)
         {
             Write($"Adapter_DeviceConnectionError {e.Device.Id.ToHexBleAddress()} with name: {e.Device.Name}");
-        }
-
-        private void Adapter_DeviceConnectionLost(object? sender, Plugin.BLE.Abstractions.EventArgs.DeviceErrorEventArgs e)
-        {
-            Write($"Adapter_DeviceConnectionLost {e.Device.Id.ToHexBleAddress()} with name: {e.Device.Name}");
         }
 
         private void Adapter_DeviceDisconnected(object? sender, Plugin.BLE.Abstractions.EventArgs.DeviceEventArgs e)
@@ -163,38 +161,68 @@ namespace BLE.Client.WinConsole
             }
         }
 
-        public async Task Connect_ConnectionLost_Connect()
+        private void ConsoleKeyReader()
+        {
+            while (consoleKey != ConsoleKey.Escape)
+            {
+                consoleKey = Console.ReadKey().Key;
+            }
+            Write("Escape key pressed - stopping...");
+            escKeyCancellationTokenSource.Cancel();
+        }
+
+        private async Task ConnectWorker(Guid id)
+        {
+            while (consoleKey != ConsoleKey.Escape)
+            {
+                try
+                {
+                    Write("Trying to connect to device (Escape key to abort)");
+                    reconnectDevice = await Adapter.ConnectToKnownDeviceAsync(id, cancellationToken: escKeyCancellationTokenSource.Token);
+                    Write("Reading all services and characteristics");
+                    var services = await reconnectDevice.GetServicesAsync();
+                    List<ICharacteristic> characteristics = new List<ICharacteristic>();
+                    foreach (var service in services)
+                    {
+                        var newcharacteristics = await service.GetCharacteristicsAsync();
+                        characteristics.AddRange(newcharacteristics);
+                    }
+                    await Task.Delay(1000);
+                    Write(new string('-', 80));
+                    Write("Connected successfully!");
+                    Write("To test connection lost: Move the device out of range / power off the device");
+                    Write(new string('-', 80));
+                    break;
+                }
+                catch
+                {
+                }
+            }
+        }
+
+        public async Task Connect_ConnectionLost_Reconnect()
         {
             string bleaddress = BleAddressSelector.GetBleAddress();
             var id = bleaddress.ToBleDeviceGuid();
-            var connectParameters = new ConnectParameters(connectionParameterSet: ConnectionParameterSet.Balanced);
-            ConsoleKey consoleKey = ConsoleKey.None;
-            using (IDevice dev = await Adapter.ConnectToKnownDeviceAsync(id, connectParameters))
+            var consoleReaderTask = new Task(ConsoleKeyReader);
+            consoleReaderTask.Start();
+            await ConnectWorker(id);
+            consoleReaderTask.Wait();
+        }
+
+        private async void Adapter_DeviceConnectionLost(object? sender, Plugin.BLE.Abstractions.EventArgs.DeviceErrorEventArgs e)
+        {
+            Write($"Adapter_DeviceConnectionLost {e.Device.Id.ToHexBleAddress()} with name: {e.Device.Name}");
+            if (reconnectDevice is not null && reconnectDevice.Id == e.Device.Id)
             {
-                while (consoleKey != ConsoleKey.Escape)
-                {
-                    Write("Reading services");
-                    var services = await dev.GetServicesAsync();
-                    List<ICharacteristic> charlist = new List<ICharacteristic>();
-                    foreach (var service in services)
-                    {
-                        var characteristics = await service.GetCharacteristicsAsync();
-                        charlist.AddRange(characteristics);
-                    }
-                    await Task.Delay(1000);
-                    Console.WriteLine(new string('-', 80));
-                    Console.WriteLine("Now powercycle the device... Hit any key when the device is booted up again (Escape to quit)");
-                    Console.WriteLine(new string('-', 80));
-                    consoleKey = Console.ReadKey().Key;
-                    await Adapter.ConnectToDeviceAsync(dev, connectParameters);
-                    Write("Waiting 3 secs");
-                    await Task.Delay(3000);
-                    foreach (var service in services)
-                    {
-                        service.Dispose();
-                    }
-                    charlist.Clear();
-                }
+                reconnectDevice.Dispose();
+                reconnectDevice = null;
+                await Task.Delay(1000);
+                Write(new string('-', 80));
+                Write("Lost connection!");
+                Write("To test reconnect: Move the device back in range / power on the device");
+                Write(new string('-', 80));
+                _ = ConnectWorker(e.Device.Id);
             }
         }
 
@@ -231,7 +259,7 @@ namespace BLE.Client.WinConsole
         public Task GetBondedDevices()
         {
             int idx = 0;
-            foreach(var dev in Adapter.BondedDevices)
+            foreach (var dev in Adapter.BondedDevices)
             {
                 Write($"{idx++} Bonded device: {dev.Name} : {dev.Id}");
             }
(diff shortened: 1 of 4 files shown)


janusw:
> To fix: #826

@CodeMusher Does this fix the problem for you?

CodeMusher:
@janusw I’ll try to check that tomorrow!

CodeMusher:
@janusw @AskBojesen 
Now I've tried to test the solution and it works much better now. 

For some reason the line below in the Adapter.cs for Windows throws System.InvalidCastException 
var conpar = nativeDevice.GetConnectionParameters();
It gives me the following StackTrace
 at WinRT.IObjectReference.As[T](Guid iid)
   at Windows.Devices.Bluetooth.BluetoothLEDevice.Make___objRef_global__Windows_Devices_Bluetooth_IBluetoothLEDevice6()
   at Windows.Devices.Bluetooth.BluetoothLEDevice.get__objRef_global__Windows_Devices_Bluetooth_IBluetoothLEDevice6()
   at Windows.Devices.Bluetooth.BluetoothLEDevice.GetConnectionParameters()
   at Plugin.BLE.Windows.Adapter.Device_ConnectionStatusChanged(BluetoothLEDevice nativeDevice, Object args) in C:\git\GitHub\dotnet-bluetooth-le\Source\Plugin.BLE\Windows\Adapter.cs:line 139
   at WinRT._EventSource_global__Windows_Foundation_TypedEventHandler_global__Windows_Devices_Bluetooth_BluetoothLEDevice__object_.EventState.<GetEventInvoke>b__1_0(BluetoothLEDevice sender, Object args)
   at ABI.Windows.Foundation.TypedEventHandler`2.Do_Abi_Invoke[TSenderAbi,TResultAbi](Void* thisPtr, TSenderAbi sender, TResultAbi args)
I needed to catch that exception to be able to continue. Strange, I don't think that thing isn't changed recently

axa88:
I dont use the console demo or any other abstraction over the base library so perhaps my feedback isnt useful here, but...

Ive tested this pull and the problem i see with it is _after_ the first Connect/Disconnect cycle, a subscription to either the Adapter.DeviceConnected and Adapater.DeviceDisconnected events will be lost. 
Meaning while subsequent calls to Connect will connect, and subsequent calls to Disconnect will disconnect a device, subscription to their associated events are no longer called.

CodeMusher:
@axa88 I’ve not tested in the real application either, but if it is like you are describing it regarding the events then it’s not good.

CodeMusher:
Now I've tried the PR a step up in the abstraction chain, that is my test application. It seems like I get both Adapter.DeviceConnected and Adapter.DeviceConnectionLost when I power cycle my peripheral. In my setup it seems to work!

axa88:
@CodeMusher 
> Now I've tried the PR a step up in the abstraction chain, that is my test application. It seems like I get both Adapter.DeviceConnected and Adapter.DeviceConnectionLost when I power cycle my peripheral. In my setup it seems to work!

Are you able to get `Adapter.DeviceConnected` _after_ a subsequent `Adapter.DeviceDisconnected`?

AskBojesen:
> @janusw @AskBojesen Now I've tried to test the solution and it works much better now.
> 
> For some reason the line below in the Adapter.cs for Windows throws System.InvalidCastException var conpar = nativeDevice.GetConnectionParameters(); It gives me the following StackTrace at WinRT.IObjectReference.As[T](Guid iid) at Windows.Devices.Bluetooth.BluetoothLEDevice.Make___objRef_global__Windows_Devices_Bluetooth_IBluetoothLEDevice6() at Windows.Devices.Bluetooth.BluetoothLEDevice.get__objRef_global__Windows_Devices_Bluetooth_IBluetoothLEDevice6() at Windows.Devices.Bluetooth.BluetoothLEDevice.GetConnectionParameters() at Plugin.BLE.Windows.Adapter.Device_ConnectionStatusChanged(BluetoothLEDevice nativeDevice, Object args) in C:\git\GitHub\dotnet-bluetooth-le\Source\Plugin.BLE\Windows\Adapter.cs:line 139 at WinRT._EventSource_global__Windows_Foundation_TypedEventHandler_global__Windows_Devices_Bluetooth_BluetoothLEDevice__object_.EventState.b__1_0(BluetoothLEDevice sender, Object args) at ABI.Windows.Foundation.TypedEventHandler`2.Do_Abi_Invoke[TSenderAbi,TResultAbi](Void* thisPtr, TSenderAbi sender, TResultAbi args) I needed to catch that exception to be able to continue. Strange, I don't think that thing isn't changed recently

The  code has been in there for a while - but it is new to 3.1 vs 3.0.

It is not good if exception is thrown for the GetConnectionParameters(), because then the HandleConnectedDevice(connectedDevice) is not called.

image

Is this consequently you get this exception?

We could make a try'n'catch in the code when calling GetConnectionParameters

janusw:
> For some reason the line below in the Adapter.cs for Windows throws System.InvalidCastException
> var conpar = nativeDevice.GetConnectionParameters();

Huh, that line as such looks rather innocent, and I don't really see why it would generate an InvalidCastException.

However `BluetoothLEDevice.GetConnectionParameters` is only available in Windows 11 (build 10.0.22000.0), and is guarded by a corresponding compile-time check. But I suspect we're missing a proper runtime API version check.

@CodeMusher Two questions:
1) Which Windows version are you targeting in your application?
2) Which Windows version are you actually running it on?

CodeMusher:
@janusw You're right that it must be the runtime that messes it up

1. Which Windows version are you targeting in your application?
I target net8.0-windows10.0.22621.0 as that is what the WinConsole sample is using

2. Which Windows version are you actually running it on?
My laptop only runs 10.0.19045 so that must be what causing the problem.

CodeMusher:
The tests I performed yesterday were power cycling the peripheral. These tests gave me
Adapter.DeviceConnected
Adapter.DeviceConnectionLost. 

Today I tried to initiate disconnect from the app instead. I thought that would invoke a Adapter.DeviceDisconnected. 
But it always invokes Adapter.DeviceConnectionLost.
Is that really what I should expect?

AskBojesen:
> @janusw You're right that it must be the runtime that messes it up
> 
> 1. Which Windows version are you targeting in your application?
>    I target net8.0-windows10.0.22621.0 as that is what the WinConsole sample is using
> 2. Which Windows version are you actually running it on?
>    My laptop only runs 10.0.19045 so that must be what causing the problem.

you need runtime check if running on Windows 10, or you must use another target than 10.0.22621
Unfortunately Windows has no check for runtime - you simply get exceptions has you saw!

I have made a runtime check as custom action for the installer for our desktop app to avoid running on eg. 10.0.19045

public static int CurrentBuild()
{
    var reg = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\Windows NT\CurrentVersion");
    var currentBuildStr = (string)reg.GetValue("CurrentBuild");
    var currentBuild = int.Parse(currentBuildStr);
    return currentBuild;
}

public override void Install(IDictionary stateSaver)
{
    // Environment.OSVersion.Version.Build does not work on .Net framework
    //int currentBuild = Environment.OSVersion.Version.Build;
    int currentBuild = CurrentBuild();
    LogText($"CurrentBuild = {currentBuild}");
    if (currentBuild < 22000)
    {
        StringBuilder sb = new StringBuilder();
        sb.AppendLine("ILab can only run on Windows 11 and higher")
                  .AppendLine()
                  .AppendLine($"CurrentBuild is {currentBuild}, which must be higher or equal to 22000");
        throw new NotSupportedException(sb.ToString());
    }
    base.Install(stateSaver);
}

AskBojesen:
I get Adapter.DeviceDisconnected when testing with the winconsole client...

See the first lines in the attached demolog.txt

Running: Connect -> Disconnect
```
18.10.05.834 Plugin.BLE - ConnectToDeviceNativeAsync 8C4B14C9C43A Named: Egoo RD123456 Connected: Disconnected
18.10.05.835 Plugin.BLE - RequestPreferredConnectionParameters(ThroughputOptimized): Success
18.10.06.090 Plugin.BLE - Device_ConnectionStatusChanged 8C4B14C9C43A Egoo RD123456 Connected
18.10.06.090 Plugin.BLE - Connected with Latency = 0, Interval = 12, Timeout = 200
18.10.06.090       DEMO - Adapter_DeviceConnected 8C4B14C9C43A with name:  Egoo RD123456
18.10.06.090 Plugin.BLE - ConnectToDeviceAsync Connected: 00000000-0000-0000-0000-8c4b14c9c43a Egoo RD123456
18.10.06.090       DEMO - Waiting 5 secs
18.10.06.090 Plugin.BLE - GattSession_SessionStatusChanged: Active
18.10.06.253 Plugin.BLE - GattSession_MaxPduSizeChanged: 517
18.10.11.085       DEMO - Disconnecting
18.10.11.085 Plugin.BLE - DisconnectDeviceNative from device with ID:  8C4B14C9C43A
18.10.14.191 Plugin.BLE - Device_ConnectionStatusChanged 8C4B14C9C43A Egoo RD123456 Disconnected
**18.10.14.191 Plugin.BLE - DisconnectedPeripheral by user: Egoo RD123456
18.10.14.191       DEMO - Adapter_DeviceDisconnected 8C4B14C9C43A with name: Egoo RD123456**
18.10.14.191 Plugin.BLE - DisconnectAsync Disconnected: 00000000-0000-0000-0000-8c4b14c9c43a Egoo RD123456
18.10.14.191       DEMO - Test_Connect_Disconnect done
```

CodeMusher:
@AskBojesen > you need runtime check if running on Windows 10, or you must use another target than 10.0.22621 Unfortunately Windows has no check for runtime - you simply get exceptions has you saw!
> 
> I have made a runtime check as custom action for the installer for our desktop app to avoid running on eg. 10.0.19045
> 

Ok, I'll keep that in mind, missed that one.

janusw:
> > @janusw You're right that it must be the runtime that messes it up
> > 
> > 1. Which Windows version are you targeting in your application?
> >    I target net8.0-windows10.0.22621.0 as that is what the WinConsole sample is using
> > 2. Which Windows version are you actually running it on?
> >    My laptop only runs 10.0.19045 so that must be what causing the problem.

Thanks for the info. This indeed confirms my suspicion.

> you need runtime check if running on Windows 10, or you must use another target than 10.0.22621 Unfortunately Windows has no check for runtime - you simply get exceptions has you saw!

Well, I think the library should better guard the version-dependent code by a runtime check and only execute it if it's supported (see https://learn.microsoft.com/en-us/windows/uwp/debug-test-perf/version-adaptive-apps).

For `BluetoothLEDevice.GetConnection` one needs to check for `UniversalApiContract` v14 AFAICS.

CodeMusher:
@AskBojesen > Disconnecting from app - which app on which platform?

It is my test app running on top of your library on Windows. I've checked the old code and it seems like it does something else than how it is done in the WinConsole. 
So I believe the issue is on my side this time.

AskBojesen:
> > > @janusw You're right that it must be the runtime that messes it up
> > > 
> > > 1. Which Windows version are you targeting in your application?
> > >    I target net8.0-windows10.0.22621.0 as that is what the WinConsole sample is using
> > > 2. Which Windows version are you actually running it on?
> > >    My laptop only runs 10.0.19045 so that must be what causing the problem.
> 
> Thanks for the info. This indeed confirms my suspicion.
> 
> > you need runtime check if running on Windows 10, or you must use another target than 10.0.22621 Unfortunately Windows has no check for runtime - you simply get exceptions has you saw!
> 
> Well, I think the library should better guard the version-dependent code by a runtime check and only execute it if it's supported (see https://learn.microsoft.com/en-us/windows/uwp/debug-test-perf/version-adaptive-apps).
> 
> For `BluetoothLEDevice.GetConnection` one needs to check for `UniversalApiContract` v14 AFAICS.

I agree - that it is dangerous as it is now - I got a nasty surprise myself trying to run the desktop app on Windows 10.
But @janusw : Can we keep focus on this PR for the original  connection lost issue - and create another PR for Windows Runtime awareness ?

AskBojesen:
@CodeMusher , @axa88 : Can you please confirm, that the original issue for this PR is solved?

Original issue: not connection lost of lost without client request

AskBojesen:
Note: open issues not fixed in this PR:

- https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/issues/852
- https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/issues/850
- https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/issues/788

When thios PR #849 is approved and merged, then I will look into the three issues above

axa88:
> @CodeMusher , @axa88 : Can you please confirm, that the original issue for this PR is solved?
> 
> Original issue: not connection lost of lost without client request

@AskBojesen 
Im unable to confirm anything regarding this issue.  With this build it doesn't take much more than 2 connect/disconnect cycles before my device is left in an unusable state which will no longer connect/disconnect as reported elsewhere.

I can report the first connect/disconnect/lost cycle does appear to work.  But just not anything past that.

CodeMusher:
I believe my initial issue is solved by this.

janusw (approved):
> I believe my initial issue is solved by this.

Alright, then let's merge it. The modification of the library itself is rather simple, after all, and ok with me.

Issue #852, opened 2024-05-14
Windows: Check OS version runtime


AskBojesen:
Windows: Check OS version number.

BLE version = 3.1.0-rc1

Currently a runtime on the Windows client needs > 22000
Eg. to call this: https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothlepreferredconnectionparameters
Unfortunately Windows allow executing code linked to > 22000 - only at runtime calling new functionality causes error.

1. Run on a Windows 10
2. Connect to an instrument

# Expected behavior
Connected without errors

# Actual behavior
Cannot connect

smsissuechecker:
Hi @AskBojesen,

I'm the friendly issue checker.
It seems like (75.00 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

If you think it's fine to make an exception, just ignore this message.
**But if you think it was a mistake to delete the template, please close the issue and create a new one.**

Thanks!

janusw:
I think this is a regression, which has been introduced in 3.1.0-beta.2 via #801. It should be fixed before we release the final version 3.1.0.

janusw:
See also the discussion in #849 ...

AskBojesen:
Ok, I will make a draft PR today to fix this

AskBojesen:
PR: https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/pull/853

Issue #847, opened 2024-05-06
Need to be able to register for notification from characteristics that is only INDICATE


CodeMusher:
To help us fix your issue, please provide the information in the below template. If something causes a crash, provide as much information as you can gather.
Just imagine: we do not know what you are doing! 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.

## Steps to reproduce

1. Try to use a peripheral with a characteristic with only INDICATE

2.

3.

## Expected behavior
Registration for updates should succeed

## Actual behavior
Registration is not performed. The Windows implementation of the StartUpdatesNativeAsync only register in case of Notify.
It could be solved this way
            GattWriteResult result;
            if (NativeCharacteristic.CharacteristicProperties.HasFlag(GattCharacteristicProperties.Indicate))
                result = await NativeCharacteristic.WriteClientCharacteristicConfigurationDescriptorWithResultAsync(GattClientCharacteristicConfigurationDescriptorValue.Indicate);
            else
                result = await NativeCharacteristic.WriteClientCharacteristicConfigurationDescriptorWithResultAsync(GattClientCharacteristicConfigurationDescriptorValue.Notify);

### Crashlog

If something causes an exception paste full stack trace + Exception here

## Configuration

**Version of the Plugin:** e.g. 1.0.0 / 1.0.1-alpha3
3.1.0-rc.1

**Platform:** e.g. iOS 10.1 / Android 4.4 / ... (including version!!! e.g. Android 5.1 / i0S 10)
UWP/Windows 10

**Device:** e.g. HTC Sensation /i Phone 7 ...

smsissuechecker:
Hi @CodeMusher,

I'm the friendly issue checker.
Thanks for using the issue template :star2:
I appreciate it very much. I'm sure, the maintainers of this repository will answer, soon.

janusw:
> ## Actual behavior
> 
[...]

PR #853, opened 2024-05-16
Windows: Check for OS build number version Runtime

AskBojesen:
To fix https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/issues/852

I have also added an installer project to easily test on another computer.
Eg. I have tested the winconsole on a Windows 10 with build number = 19041


janusw:
@CodeMusher You reported in #849 that there are `InvalidCastException` when targeting Windows builds > 22k, but running on builds < 22k. Can you confirm that this PR fixes those exceptions?

Source/Plugin.BLE/Windows/BleImplementation.cs
```diff
@@ -20,6 +20,26 @@ public class BleImplementation : BleImplementationBase
...
+        {
+            get
+            {
+                if (runtimeBuildNumber is null)
+                {
+                    runtimeBuildNumber = Environment.OSVersion.Version.Build;
+                }
+                return runtimeBuildNumber.Value;
+            }
+        }
```
janusw, line 41: To be honest, I don't see much of a point in adding this property.
Logically, it is not related to BLE at all, but is simply a general property of the OS.
Practically, one could just query `Environment.OSVersion.Version.Build` directly, which is simpler (and IMHO preferable, unless there is a good reason for adding a new property in BleImplementation).

AskBojesen: The reason is that I were inspired from my own project, where I need a solution for .net framework.
beacuse Environment.OSVersion.Version.Build does not work on .Net framework
So I needed this special method below.,
I agree to simplify now we can use Environment.OSVersion.Version.Build
```
  public static int CurrentBuild()
  {
      var reg = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\Windows NT\CurrentVersion");
      var currentBuildStr = (string)reg.GetValue("CurrentBuild");
      var currentBuild = int.Parse(currentBuildStr);
      return currentBuild;
  }
```

CodeMusher:
Yes, it does!16 maj 2024 kl. 21:31 skrev Janus Weil ***@***.***>:﻿
@CodeMusher You reported in #849 that there are InvalidCastException when targeting Windows builds > 22k, but running on builds < 22k. Can you confirm that this PR fixes those exceptions?

Source/Plugin.BLE/Windows/BleImplementation.cs
janusw: > The reason is that I were inspired from my own project, where I need a solution for .net framework. beacuse Environment.OSVersion.Version.Build does not work on .Net framework So I needed this special method below.

Thanks for the comment. Note: For Android version checks in Plugin.BLE we also use a slightly different mechanism on .NET 6 and above, see e.g. here:

https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/c894e46d7b51c3ea4debb10320483b08eea17a32/Source/Plugin.BLE/Android/Adapter.cs#L211-L215

IIRC, then `Build.VERSION.SdkInt` should also work with the latest .NET versions, but `OperatingSystem.IsAndroidVersionAtLeast` is better at removing deprecation warnings in such conditional code.

> I agree to simplify now we can use Environment.OSVersion.Version.Build

Thanks for updating this. Looks much better to me now :)

janusw (approved):
So, yeah, this is ok with me (with the latest changes and the confirmation that it fixes the issue).

janusw:
I'd tend toward releasing 3.1.0 now (today or tomorrow). I think the most urgent things are fixed and further bugfixes can still be added in subsequent 3.1.x releases.

@AskBojesen Is this ok with you, or do you want anything else to be included in 3.1.0? Do we need another rc?

AskBojesen:
> I'd tend toward releasing 3.1.0 now (today or tomorrow). I think the most urgent things are fixed and further bugfixes can still be added in subsequent 3.1.x releases.
> 
> @AskBojesen Is this ok with you, or do you want anything else to be included in 3.1.0? Do we need another rc?

No thanks! I tried to reproduce disconnect hangs sometimes on Windows. But could not - did 200 connect/disconnect on two different devices.

Looking forward to the 3.1.0 release :-)

CodeMusher:
Hi, would it be possible to get a fix for my issue regarding INDICATE characteristic on Windows into the release?Skickat från min iPhone19 maj 2024 kl. 11:23 skrev Ask Magnussen Bojesen ***@***.***>:﻿

I'd tend toward releasing 3.1.0 now (today or tomorrow). I think the most urgent things are fixed and further bugfixes can still be added in subsequent 3.1.x releases.
@AskBojesen Is this ok with you, or do you want anything else to be included in 3.1.0? Do we need another rc?

No thanks! I tried to reproduce disconnect hangs sometimes on Windows. But could not - did 200 connect/disconnect on two different devices.
Looking forward to the 3.1.0 release :-)

CodeMusher:
Yes, #847 is the one. I’ve written a solution in the issue that works for my use case. @AskBojesen do you think that is acceptable or could a characteristic have both INDICATE and NOTIFY?

janusw:
> @AskBojesen do you think that is acceptable or could a characteristic have both INDICATE and NOTIFY?

Please continue this discussion in #847 (or the PR you'll open to fix it). It is highly off-topic here ;)

PR #855, opened 2024-05-17
fix #854 - Bluetooth stack breaks when trying connect on phone boot

### LINKED ISSUE #854: Bluetooth stack breaks when trying connect on phone boot
Some devices load cache of previously connected devices **slowly**.
When you try to ConnectToDeviceAsync while loading the cache, the bluetooth stack of the device completely breaks.
Before connecting, you should check the device type to avoid this break. Check Source/Plugin.BLE/Android/Adapter.cs ConnectToKnownDeviceNativeAsync . BluetoothDeviceType must be Le or Dual not Unknown!

```
public override async Task<IDevice> ConnectToKnownDeviceNativeAsync(Guid deviceGuid, ConnectParameters connectParameters = default(ConnectParameters), CancellationToken cancellationToken = default(CancellationToken))
{
    var macBytes = deviceGuid.ToByteArray().Skip(10).Take(6).ToArray();
    var nativeDevice = _bluetoothAdapter.GetRemoteDevice(macBytes);
    if (nativeDevice == null
      || !(BluetoothDeviceType.Le == nativeDevice.Type || BluetoothDeviceType.Dual == nativeDevice.Type) /* Check the device type */ )
        throw new Abstractions.Exceptions.DeviceConnectionException(deviceGuid, "", $"[Adapter] Device {deviceGuid} not found.");
    var device = new Device(this, nativeDevice, null);

    await ConnectToDeviceAsync(device, connectParameters, cancellationToken);
    return device;
}
```

## Steps to reproduce

1. Reboot device
2. wait 2 or 3 minutes
3. Start any your app and connect to any BLE device (exsamle Source/BLE.Client : DeviceListPage.xaml button "Connect to previous" calls Adapter.ConnectToKnownDeviceAsync -> ConnectToKnownDeviceNativeAsync) - it is OK
4. Reboot device
5. do not wait, Start any your app and start connection to BLE device. 

## Actual behavior
Phone Bluetooth stack breaks and does not work at all until the next reboot - unable to connect any BLE devices.

## Expected behavior
Bluetooth devices connect and function normally

## Configuration
**Version of the Plugin:** all: 2 and 3
**Platform:** Android 9+
**Device:** china phones: IIIF150 https://www.iiif150.com/

alex3696:
see issue #854 - Bluetooth stack breaks when trying to connect on phone boot


Source/Plugin.BLE/Android/Adapter.cs
```diff
@@ -304,7 +304,7 @@ public override async Task<IDevice> ConnectToKnownDeviceNativeAsync(Guid deviceG
         {
             var macBytes = deviceGuid.ToByteArray().Skip(10).Take(6).ToArray();
             var nativeDevice = _bluetoothAdapter.GetRemoteDevice(macBytes);
-            if (nativeDevice == null)
+            if (nativeDevice == null || !(BluetoothDeviceType.Le == nativeDevice.Type || BluetoothDeviceType.Dual == nativeDevice.Type))
                 throw new Abstractions.Exceptions.DeviceConnectionException(deviceGuid,"", $"[Adapter] Device {deviceGuid} not found.");
             var device = new Device(this, nativeDevice, null);
```
janusw, line 309: Thanks for the contribution! Looks pretty good :)

Two minor comments/suggestions:
1. How about introducing a small auxiliary method in the `Adapter` class? (Apart from this place, it could also be used in `GetSystemConnectedOrPairedDevices` and `GetBondedDevices`.)
```
        static bool SupportsBLE(BluetoothDevice d)
        {
            return d.Type == BluetoothDeviceType.Le || d.Type == BluetoothDeviceType.Dual;
        }
```
2. I think the error message in the exception is not fully accurate for the new case you're adding (the device with the given GUID exists, but just does not support BLE). It might be better to generate an exception with a slightly different error message. Maybe something like this: "[Adapter] Device {deviceGuid} does not support BLE."

alex3696 (commented):
ready

alex3696 (commented):
I like the class extension snippets.

janusw:
> I like the class extension snippets.

Yes, that's even better :+1:

janusw (approved):
LGTM by now. Thanks for the contribution!

PR #858, opened 2024-05-19
Fixed StartUpdatesNativeAsync for indicate

### LINKED ISSUE #847: Need to be able to register for notification from characteristics that is only INDICATE

AskBojesen:
To fix #847

### DIFF
--- a/Source/BLE.Client/BLE.Client.WinConsole/PluginDemos.cs
+++ b/Source/BLE.Client/BLE.Client.WinConsole/PluginDemos.cs
@@ -103,7 +103,7 @@ namespace BLE.Client.WinConsole
             string bleaddress = BleAddressSelector.GetBleAddress();
             var id = bleaddress.ToBleDeviceGuid();
             var connectParameters = new ConnectParameters(connectionParameterSet: ConnectionParameterSet.Balanced);
-            new Task(ConsoleKeyReader).Start();            
+            new Task(ConsoleKeyReader).Start();
             using (IDevice dev = await Adapter.ConnectToKnownDeviceAsync(id, connectParameters))
             {
                 int count = 1;
@@ -151,7 +151,7 @@ namespace BLE.Client.WinConsole
             var connectParameters = new ConnectParameters(connectionParameterSet: ConnectionParameterSet.Balanced);
             new Task(ConsoleKeyReader).Start();
             int count = 1;
-            while (true)  
+            while (true)
             {
                 await Task.Delay(100);
                 Write($"---------------- {count++} ------- (Esc to stop) ------");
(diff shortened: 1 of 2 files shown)


CodeMusher:
@janusw, I haven't got the possibility to verify it today but it looks almost identical to the solution I provided when opening the issue so I guess it must.

janusw:
> @janusw, I haven't got the possibility to verify it today

No problem. When will you be able to verify it?

> but it looks almost identical to the solution I provided when opening the issue

Yes, it's very close, but not identical.

> so I guess it must.

Sorry, but guessing is not good enough here. Apparently @AskBojesen was not able to reproduce the issue and test the fix, and I'm not going to accept an untested PR at this stage (i.e. directly before a stable release).

CodeMusher:
I'll try to test it tomorrow

AskBojesen:
BTW: I have tested using my "normal" device with Notity in char descriptors  - and that still works as expected.

CodeMusher:
Now I have tested with two different devices. One with characteristics with only INDICATE and one with a mixed population of characteristics, some INDICATE and some NOTIFY. 
It works as expected!

janusw (approved):
Since both of you have tested it and code-wise it looks very clean, this is ok with me. Thanks for fixing this so quickly, @AskBojesen!

Issue #824, opened 2024-03-06
WriteAsync does not use the CancellationToken that is given to it


minno726:
## Steps to reproduce

1. Call `WriteAsync` by passing a `CancellationToken` to it.

2. Before the call completes, cancel the token. This is easiest to do by passing a very small timeout in the constructor for `CancellationTokenSource`.

## Expected behavior
The operation should be canceled.

## Actual behavior
The operation ignores the cancellation.

## Configuration

**Version of the Plugin:** 3.0.0

**Platform:** Observed on iOS due to #785, but the problem is in shared code.

The source of the problem is here:

https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/b179852efef3402a74c3c513dfce004c9c158077/Source/Plugin.BLE/Shared/DescriptorBase.cs#L68-L76

The code does not use the cancellation token that was passed in as a function parameter. See the documentation on how to use cancellation tokens here: https://learn.microsoft.com/en-us/dotnet/standard/threading/cancellation-in-managed-threads, or remove the parameter if cancellation isn't practical to implement.

This is related to #785, since it prevents one workaround for that issue from working.

smsissuechecker:
Hi @minno726,

I'm the friendly issue checker.
It seems like (25.00 %) you haven't used our issue template :cry: I think it is very frustrating for the repository owners, if you ignore them.

If you think it's fine to make an exception, just ignore this message.
**But if you think it was a mistake to delete the template, please close the issue and create a new one.**

Thanks!

ygl-rg:
```c#
public static async Task<int> SendBytes(this ICharacteristic handle, byte[] data, CancellationToken token)
    {
        var cs = new TaskCompletionSource();
        await using(token.Register(() =>
        {
            cs.TrySetResult();
        }));
        var t1 = handle.WriteAsync(data, token);
        var t = await Task.WhenAny(cs.Task, t1);
        if (t == cs.Task)
            throw new TimeoutException("Send bytes timeout");
        return await t1;
    }
```

I use the above method for cancellation.

bcaceiro:
Hi @janusw , are you planning on addressing this issue? We have a temp workaround, however would be best if the cancellation would respect the given CancellationToken

janusw:
> Hi @janusw , are you planning on addressing this issue?

[...]

Issue #785, opened 2023-12-05
[iOS] Characteristic.WriteAsync() hangs


softeip:
I have a MAUI app which needs to monitor realtime data from BLE device.
Usually I send ~20 bytes to the device and expect to receive ~300-500 bytes back. Each packet gets sliced by 155 bytes per notification, so smartphone sends one notification in this case and receives multiple. What I noticed is that the app stops sending and receiving notifications after some random time (usually after 30-45 mins, but sometimes it hangs after 5-10 mins). 
I've tried to add more logging and spent hours on debugging and now I see that Characteristic.WriteAsync() call hangs for unknown reasons. I tried to pass cancellation token to its params and cancel it after ~200 ms - no reaction. Also I ensured that I call the method on the main thread.

About characteristic - WriteType is set to WithoutResponse. Trying to write data WithResponse throws an error saying that it is not supported by device or sth like that.

Sometimes I also gets an exception saying that my device disconnected but I don't know if it is related to my issue:
Device *** disconnected while writing characteristic with ***
**at** Plugin.BLE.Abstractions.Utils.TaskBuilder.<SafeEnqueueAndExecute>d__11`1[[System.Int32, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].MoveNext()
**at** Plugin.BLE.Abstractions.Utils.TaskBuilder.<FromEvent>d__9`3[[System.Int32, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.EventHandler, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.EventHandler`1[[CoreBluetooth.CBPeripheralErrorEventArgs, Microsoft.iOS, Version=16.4.0.0, Culture=neutral, PublicKeyToken=84e04ff9cfb79065]], System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].MoveNext()
**at** Plugin.BLE.Abstractions.CharacteristicBase`1.<WriteAsync>d__34[[CoreBluetooth.CBCharacteristic, Microsoft.iOS, Version=16.4.0.0, Culture=neutral, PublicKeyToken=84e04ff9cfb79065]].MoveNext()

## Configuration

[...]

PR #880, opened 2024-08-09
recognize cancellation tokens for async methods

srruu:
Some of the cancellation-tokens of the public api where ignored (e.g. WriteAsync, ReadAsync, ...) which is fixed in this PR.
There were also some public async methods without a cancellation token parameter. They now also have a cancellation token parameter.
This solves #824 and should prevent any "hanging" tasks like descriped in #785 (this will not fix the root cause).


janusw:
Thanks for the PR!

Unfortunately the Mac build is failing, but I don't think this is caused by your changes. The relevant warnings/errors are:

```
warning MT0079: The recommended Xcode version for Microsoft.iOS 17.5.8018 is Xcode 15.4 or later.
The current Xcode version (found in /Applications/Xcode_15.2.app/Contents/Developer) is 15.2.
[/Users/runner/work/dotnet-bluetooth-le/dotnet-bluetooth-le/Source/BLE.Client/BLE.Client.Maui/BLE.Client.Maui.csproj::TargetFramework=net8.0-ios]

error MT0180: This version of Microsoft.iOS requires the iOS 17.5 SDK (shipped with Xcode 15.4).
Either upgrade Xcode to get the required header files or set the managed linker behaviour to
Link Framework SDKs Only in your project's iOS Build Options > Linker Behavior (to try to avoid the new APIs).
[/Users/runner/work/dotnet-bluetooth-le/dotnet-bluetooth-le/Source/BLE.Client/BLE.Client.Maui/BLE.Client.Maui.csproj::TargetFramework=net8.0-ios]
```

Sounds like we need to update the XCode version used in the GHA build on Mac. I can look into that.

janusw:
> Sounds like we need to update the XCode version used in the GHA build on Mac. I can look into that.

... and for that we probably need to update the build host to MacOS 14 (it is currently using 13.6.8).

See:
* https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md
* https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md

janusw:
Fixed the build problems via #881 & rebased this one on top of it  ➡️  all builds succeed ✅ 🥳

Source/Plugin.BLE/Shared/AdapterBase.cs
```diff
@@ -226,7 +226,7 @@ public async Task ConnectToDeviceAsync(IDevice device, ConnectParameters connect
         /// Disconnects from the <paramref name="device"/>.
         /// </summary>
         /// <param name="device">Device to connect from.</param>
-        public Task DisconnectDeviceAsync(IDevice device)
+        public Task DisconnectDeviceAsync(IDevice device, CancellationToken cancellationToken = default)
```
janusw, line 229: The GHA build shows a warning here:
```
Parameter 'cancellationToken' has no matching param tag in the XML comment for 'AdapterBase.DisconnectDeviceAsync(IDevice, CancellationToken)' (but other parameters do)
```
@srruu Could you please fix this warning by documenting the `cancellationToken` parameter in the comment above the method?

Source/Plugin.BLE/Shared/Contracts/IDevice.cs
```diff
@@ -77,7 +78,7 @@ public interface IDevice : IDisposable
         /// A task that represents the asynchronous read operation. The Result property will contain a boolean that indicates if the update was successful.
         /// The Task will finish after the RSSI has been updated.
         /// </returns>
-        Task<bool> UpdateRssiAsync();
+        Task<bool> UpdateRssiAsync(CancellationToken cancellationToken = default);
```
janusw, line 81: This change causes a warning on line 25 of this file (and also in line 75 of DeviceBase.cs):
```
XML comment has cref attribute 'UpdateRssiAsync()' that could not be resolved
```
I think the `cref` also needs to include the new parameter (or rather: it's type), probably like this:
```
cref="UpdateRssiAsync(CancellationToken)"
```

Source/Plugin.BLE/Shared/ServiceBase.cs
```diff
@@ -61,16 +62,16 @@ public async Task<IReadOnlyList<ICharacteristic>> GetCharacteristicsAsync()
         /// Gets the first characteristic with the Id <paramref name="id"/>. 
         /// </summary>
         /// <param name="id">The id of the searched characteristic.</param>
-        public async Task<ICharacteristic> GetCharacteristicAsync(Guid id)
+        public async Task<ICharacteristic> GetCharacteristicAsync(Guid id, CancellationToken cancellationToken = default)
```
janusw, line 65: Another warning because the new parameter is not documented:
```
Parameter 'cancellationToken' has no matching param tag in the XML comment for 'ServiceBase<TNativeService>.GetCharacteristicAsync(Guid, CancellationToken)' (but other parameters do)
```

janusw (changes requested):

janusw:
@minno726 @bcaceiro Does this fix #824 for you?
@softeip @sschoeb Does it fix #785 for you?

Could anyone of you guys test this PR and report if it works for you? You can download the nupkg from https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/actions/runs/10338642672#artifacts

ScottyLightnin:
This would be a nice win. I've been frustrated when trying to cancel a call, only for the token to be ignore.

softeip:
It's really hard to cancel token after calling ICharacteristic.WriteAsync() since it gets executed fast enough.
I could not catch any hanging issues that I had mentioned in #785 which I expected to bump into to help testing cancellation - the token would be automatically cancelled in this case by CancelAfter(10000) 

I tried passing already cancelled token and I got no exception thrown by ICharacteristic.WriteAsync() method - is it expected behavior is this case or have I done something wrong?

All I can confirm regarding `3.1.0-6-gffd5b3d` is that my app works as stable as it does with 3.1.0.

janusw:
> I tried passing already cancelled token and I got no exception thrown by ICharacteristic.WriteAsync() method - is it expected behavior is this case or have I done something wrong?

@srruu Maybe one should add a corresponding check and throw a `System.ArgumentException` for that case? What do you think about that?

Otherwise I'm happy with this PR by now (after you have address my previous review comments). 😃 


> All I can confirm regarding `3.1.0-6-gffd5b3d` is that my app works as stable as it does with 3.1.0.

Thanks for testing, @softeip ! 👍

janusw:
> > I tried passing already cancelled token and I got no exception thrown by ICharacteristic.WriteAsync() method - is it expected behavior is this case or have I done something wrong?
> 
> @srruu Maybe one should add a corresponding check and throw a `System.ArgumentException` for that case? What do you think about that?
> 
> Otherwise I'm happy with this PR by now (after you have addressed my previous review comments). 😃

@srruu Any comment here? If not, I'll just merge this as is (probably on the weekend).

dartasen:
Works perfectly on my app too 😄

janusw (approved):

janusw:
Alright, I have just merged this now (sorry for the delay).

Thanks for the contribution, @srruu !

Auto72:
@janusw 
Has this change been released to the latest NuGet package already?
When are you going to build the library adding .net9.0 ?

PR #899, opened 2024-11-08
Added CustomPairing for Windows

MartinBeckerKT:
`PairAsync ` didn't work for Windows.
See [stack overflow](https://stackoverflow.com/questions/45191412/deviceinformation-pairasync-not-working-in-wpf)

* Added CustomPairing
* added default case to PairingHandler for error traceing

### DIFF
--- a/Source/Plugin.BLE/Windows/Adapter.cs
+++ b/Source/Plugin.BLE/Windows/Adapter.cs
@@ -52,6 +52,7 @@ namespace Plugin.BLE.Windows
             DeviceInformationCustomPairing p = deviceInformation.Pairing.Custom;
             p.PairingRequested += PairingRequestedHandler;
             var result = await p.PairAsync(DevicePairingKinds.ConfirmOnly);
+            p.PairingRequested -= PairingRequestedHandler;
             Trace.Message($"BondAsync pairing result was {result.Status} with: {device.Name}: {device.Id}");
         }
 
--- a/Source/Plugin.BLE/Windows/Adapter.cs
+++ b/Source/Plugin.BLE/Windows/Adapter.cs
@@ -49,8 +49,24 @@ namespace Plugin.BLE.Windows
                 Trace.Message($"BondAsync cannot pair with: {device.Name}: {device.Id}");
                 return;
             }
-            DevicePairingResult result = await deviceInformation.Pairing.PairAsync();
+            DeviceInformationCustomPairing p = deviceInformation.Pairing.Custom;
+            p.PairingRequested += PairingRequestedHandler;
+            var result = await p.PairAsync(DevicePairingKinds.ConfirmOnly);
             Trace.Message($"BondAsync pairing result was {result.Status} with: {device.Name}: {device.Id}");
+        }
+
+        private static void PairingRequestedHandler(DeviceInformationCustomPairing sender, DevicePairingRequestedEventArgs args)
+        {
+            switch (args.PairingKind)
+            {
+                case DevicePairingKinds.ConfirmOnly:
+                    args.Accept();
+                    break;
+
+                default:
+                    Trace.Message("PairingKind " + args.PairingKind + " not supported");
+                    break;
+            }
         }
 
         protected override Task StartScanningForDevicesNativeAsync(ScanFilterOptions scanFilterOptions, bool allowDuplicatesKey, CancellationToken scanCancellationToken)


janusw:
Thanks for the contribution, @MartinBeckerKT !

@AskBojesen, could you possibly have a look at this?

janusw:
@axa88, you mentioned that you have problems with the Windows implementation. Is this one of them? Can you confirm that this fixes `Adapter.BondAsync` on Windows?

axa88:
@janusw 

alright... a quick review of the code, it appears the pull would do as indicated in most scenarios.... but as far as i understand it is not complete.

In the pairing process it is the responsibility of the _requestor_ to request the pairing modes it would like to use.  The  _requestee_ would then make the final decision on which one of those modes could be supported if any.

What this pull does, or rather allows this library to do, is continue to not allow a way for the `requestor` to request anything, only then hoping the `requestee` decides to respond with a mode that this library happens to support.  Though i understand incomplete features are common here.

But not to just bitch about the problem I did already solve it.  So Ill submit a proposal and subsequent pull to address this, as well as unify the Android signature as well while addressing a few masked concerns. 
So give me 24 hours, then if you dont hear from me your welcome to do what as you wish with this pull.

SysadmEgooHealth (approved):

AskBojesen (approved):
I approve the changes. The code there is quite similar to what I use myself.
I have tested successfully usin the wincolse demo BonAsync
15.49.30.224 Plugin.BLE - BondAsync pairing result was Paired with: Egoo 240103001: 00000000-0000-0000-0000-8c4b14814d72

(Sorry I was logged in as another github user SysadmEgooHealth... before)

axa88:
If you're going to go with this tweak, at least shouldn't the pairing challenge event handler be unsubscribed from since the result is being awaited.  I _think_ device information might be be marshaled and an unmanaged static resource.

MartinBeckerKT:
We've added the unregistration of the callback handler just to make sure no resource will leak here.

janusw:
> We should understand what this change is doing. It is making _every_ pair request on windows require the pairing device to have at least the capability to Consent. This means any device that doesn't at least support consent will now fail. 

Doesn't every pairing request have to be accepted in some way (be it programmatically or via user interaction)? I.e. every pairable device should "support" that, no?

In order to support the more advanced options from `DevicePairingKinds`, some more work will be necessary, I guess. `ConfirmOnly` is really the bare minimum, and this PR is really just a bugfix for that case IIUC. Correct?

> Also if you're going to go with this tweak, shouldn't the pairing challenge event handler at least be unsubscribed from since the result is being awaited.

This has been addressed already. Thanks, @MartinBeckerKT !

janusw (approved):
To me this PR is ok by now, and I will merge it soon, if there are no further objections.

axa88:
> > We should understand what this change is doing. It is making _every_ pair request on windows require the pairing device to have at least the capability to Consent. This means any device that doesn't at least support consent will now fail.
> 
> Doesn't every pairing request have to be accepted in some way (be it programmatically or via user interaction)? I.e. every pairable device should "support" that, no?

No, it doesn't appear to me that's how the spec is intended.  Apparently there is legacy pairing called "Just works" that will leave out the challenge response back to the requestee.
So as suggested, if a device you want to pair with only supports 'just works' and then this pull tells it 'i must get a confirmation', im betting it will break (dont have a just works device to test).  Saving grace here may be that not many are using this type of pairing anymore due to a push towards more security, let alone using this API on windows to do it.
Not suggesting you dont just accept the pull, suggesting it might well be a breaking change.

> In order to support the more advanced options from `DevicePairingKinds`, some more work will be necessary, I guess. 

Yes.
i plan to do it when i can and should address anything this pull may introduce.

> `ConfirmOnly` is really the bare minimum, and this PR is really just a bugfix for that case IIUC. Correct?

Depends on how you define minimum.
Its all semantics. Not sure if we can call a feature that was not (yet) supported a bug, and adding code to support one aspect and (possibly) break another a fix.
The guy using 'just works' today is going to have a whole different take on this...

axa88:
Finally got around to read the Bluetooth spec, found the reference explaining the pairing requirements.

There was something introduced called SSP Simple Secure Pairing, and a section which describes "Security Features of Bluetooth Low Energy".  Where it explains that "If none of the OOB, Numeric Comparison, or Passkey Entry association models are possible because of device input/output limitations, then the Just Works pairing method is used."
And with Just Works "The user is required to accept a connection without verifying the calculated value on both devices".

This means that the "Consent" method this pull indicates as minimal, would be compliant if this version of the API is deemed only compatible with SSP pairing on Windows.  And since SSP was adopted in version Bluetooth 2.1 + EDR in 2007, and apparently ironed out in version 4.1 available for 2014.  There inst likely too many devices, if any, using "Legacy Paring" with BLE making my concerns not an issue.

janusw:
Sorry for the late reply here. It seems everyone is happy with this PR by now, so I'll merge it.

PR #911, opened 2024-12-22
Unhandled exception reading BondState after Disconnect

### LINKED ISSUE #910: Windows After Device disconnect, reading Device.BondState will crash since disconnection disposes of Native Device
After Device disconnection reading `Device.BondState` property will throw since disconnection event handler calls a method called `((Device)disconnectedDevice).DisconnectInternal();` which disposes of the Native Device via `DisposeNativeDevice();`

Then when logging the exception, the application crashes because the very null Native device that threw the exception in the first place is then used in the catch clause!

may be a bit early to tell but dare i say fixing this has solved #850, #875, #876

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.

## Steps to reproduce

1. Connect then disconnect a device

2. Read its IDevice.BondState.

3.

## Expected behavior
the bondstate = NotSupported

## Actual behavior
application termination

## Configuration

**Version of the Plugin:** 3.0.1

**Platform:** Windows

**Device:** 
Qualcomm FastConnect 6900 Bluetooth Adapter Wi-Fi 6, 6E, Bluetooth 5.3

axa88:
Problem described in #910.

### DIFF
--- a/Source/Plugin.BLE/Windows/Device.cs
+++ b/Source/Plugin.BLE/Windows/Device.cs
@@ -248,19 +248,14 @@ namespace Plugin.BLE.Windows
 
         public override bool SupportsIsConnectable { get => true; }
 
-        protected override DeviceBondState GetBondState()
-        {
-            try
-            {
-                DeviceInformation deviceInformation = DeviceInformation.CreateFromIdAsync(NativeDevice.DeviceId).AsTask().Result;
-                return deviceInformation.Pairing.IsPaired ? DeviceBondState.Bonded : DeviceBondState.NotBonded;                
-            }
-            catch (Exception ex)
-            {
-                Trace.Message($"GetBondState exception for {NativeDevice.DeviceId} : {ex.Message}");
-                return DeviceBondState.NotSupported;
-            }
-        }
+		protected override DeviceBondState GetBondState()
+		{
+			if (NativeDevice == null)
+				NativeDevice = BluetoothLEDevice.FromBluetoothAddressAsync(Id.ToBleAddress()).AsTask().Result;
+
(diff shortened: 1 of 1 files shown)


janusw:
In principle this looks quite reasonable.

OTOH, one could solve the issue in a different (and possibly more appropriate) way: Since you mention in #910 that the problem occurs after calling `Device.Disconnect`, wouldn't it be better to avoid disposing the `NativeDevice` upon disconnection, instead of restoring it afterwards? Like this ...

```
--------------------- Source/Plugin.BLE/Windows/Device.cs ---------------------
index bacd124..d5aeb86 100644
@@ -219,9 +219,8 @@ namespace Plugin.BLE.Windows
         public void DisconnectInternal()
         {
             DisposeGattSession();
             ClearServices();
-            DisposeNativeDevice();
         }
 
         public override void Dispose()
         {
```

I imagine the `NativeDevice` being null could cause trouble in other situations as well (e.g. trying to reconnect after disconnecting), right?

axa88:
Frankly this pull does little to nothing to solve the underlying problem.

Strictly speaking the problem occurs just after disconnect because on windows you need to dispose of the native device in order to disconnect, thus trying to use it in this state caused the exception as the catch block actually tried to use disposed native device that threw in the first place.
This change tried to recreate instead, only that is a mistake that is repeated over and over again in the current implementation.
Do not merge.

And after assessing the situation, this is one of the main issue with the current implementation. As mentioned above it spawns off multiple copies of the native deice which then leak and prevent disconnection and subsequent reconnection.
And it fails to use the mechanisms provided by the API to maintain synchronization with the OS.

Im a few days into a rewrite im already confident ill get it right and post on my for when done.

axa88:
doesn't appear to me this issue wont be resolved from a small tweak

PR #820, opened 2024-02-29
Windows Implementation for BondAsync and BondedDevices

AskBojesen:
Implemented and tested:

- Adapter.BondAsync
- Adapter.BondedDevices

@axa88: A PR for you :-)

@janusw:  Basic implementation of  methods for Windows

PR #915, opened 2025-01-05
Fixes #914, #820 DeviceBondStateChanged now works on windows

### LINKED ISSUE #914: Windows DeviceBondStateChanged event doesnt work/isnt supported.
> `DeviceBondStateChanged`

I looked into this now - and I doubt that the DeviceBondStateChanged can be implemented for Windows.
The DeviceInformation API doesn't contain such events.
The device enumeration doesn't contain such events either.

@janusw: I think think the PR should be merged.

@axa88 : What is your use case for the DeviceBondStateChanged  event?

https://learn.microsoft.com/en-us/windows/uwp/devices-sensors/enumerate-devices#enumerate-and-watch-devices

_Originally posted by @AskBojesen in https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/issues/820#issuecomment-2018017306_

I dont see why it couldn't be

axa88:
Is it a bug? is it a feature? who knows, but its quick and easy so why not.

### DIFF
--- a/Source/Plugin.BLE/Windows/Adapter.cs
+++ b/Source/Plugin.BLE/Windows/Adapter.cs
@@ -19,16 +19,35 @@ namespace Plugin.BLE.Windows
         private readonly BluetoothAdapter _bluetoothAdapter;
         private BluetoothLEAdvertisementWatcher _bleWatcher;
 
-        /// <summary>
+		/// <summary>
         /// Registry used to store device instances for pending disconnect operations
         /// Helps to detect connection lost events.
         /// </summary>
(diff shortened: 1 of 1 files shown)


AskBojesen:
@axa88, I agree that this will work.
Personally I do not use the `DeviceBondStateChanged `event in the `interface IAdapter `
But I guess if you have an application running on Android  (Where it works) and Windows then this feature makes sense
I wonder if we could make it optional somehow - so that one could avoid having this Watcher running if not using it.
Maybe a new `public static bool` in the `public class BleImplementation` `\Plugin.BLE\Windows\BleImplementation.cs` to control this, like we have the CacheModeCharacteristicRead etc...
What do you think @janusw ?

PR #928, opened 2025-02-23
provide missing cancellation token to write characteristics method

srruu:
The Characteristics.WriteNativeAsync method was not recognizing the cancellation token in mode WriteWithoutResponse on iOS (forget this in my last pull request).


janusw (approved):

janusw:
Thanks for the contribution and sorry for the late reply!

PR #979, opened 2026-01-14
Apple: Update default GetSystemConnectedOrPaired behaviour

skwort:
Calls to `Plugin.Ble.Apple.Adapter.GetSystemConnectedOrPaired` with a null serviceUuids argument causes a `System.ArgumentNullException` in the `centralMananger.RetrieveConnectedPeripherals` call.

Update the Apple adapter method to pass an empty list when a null services list is provided, avoiding the null exception. This standardises the default behaviour across the supported platforms.

---

We noticed this inconsistency when testing our application; Apple devices would throw an exception, while the other platform wouldn't. While it's probably a reasonable expectation to wrap this in a try-catch, I think it makes more sense to just change the default behaviour to be consistent across the platforms; a null services argument only causes exceptions on Apple devices.

---

EDIT: Had some formatting issues on my local machine. Apologies for multiple force-pushes. This has been fixed now.

### DIFF
--- a/Source/Plugin.BLE/Apple/Adapter.cs
+++ b/Source/Plugin.BLE/Apple/Adapter.cs
@@ -272,6 +272,10 @@ namespace Plugin.BLE.iOS
             {
                 serviceUuids = services.Select(guid => CBUUID.FromString(guid.ToString())).ToArray();
             }
+            else
+            {
+                serviceUuids = [];
+            }
 
             var nativeDevices = _centralManager.RetrieveConnectedPeripherals(serviceUuids);
 


janusw:
Thanks for the contribution! Looks pretty straightforward :)

> We noticed this inconsistency when testing our application;

Is there any issue report relating to this problem? A quick search through the issue list didn't show anything. I'm a bit surprised that this hasn't surfaced before ... 😆

skwort:
I didn't bother making an issue for this problem; it was a relatively simple fix so I decided to just make the PR. I'm happy to create an issue with a stacktrace and link it with steps to reproduce if you think that would be valuable.

All that's need to reproduce it is to call `CrossBluetoothLE.Current.Adapter.GetSystemConnectedOrPaired()`.

Here's the trace from our iOS dev (I've modifed this a bit for privacy):

```
2026-01-12 11:29:05.207 Redacted.iOS[8696:1384680] 11:29:05.206 fail: System.ArgumentNullException: Value cannot be null. (Parameter 'serviceUUIDs')
2026-01-12 11:29:05.208 Redacted.iOS[8696:1384680] at CoreBluetooth.CBCentralManager.RetrieveConnectedPeripherals(CBUUID[] serviceUUIDs) in /Users/builder/azdo/_work/1/s/macios/src/build/dotnet/ios/generated-sources/CoreBluetooth/CBCentralManager.g.cs:line 256 at Plugin.BLE.iOS.Adapter.GetSystemConnectedOrPairedDevices(Guid[] services)
2026-01-12 11:29:05.208 Redacted.iOS[8696:1384680] at 
...
```

janusw:
> I didn't bother making an issue for this problem; it was a relatively simple fix so I decided to just make the PR. I'm happy to create an issue with a stacktrace and link it with steps to reproduce if you think that would be valuable.

No worries, I don't think it's necessary to open an issue for it (just thought we might have one already ...)

janusw (approved):
I was able to reproduce the problem on MacOS, and can confirm that this PR fixes it.

janusw:
Merged. Thanks again for the contribution!

Issue #985, opened 2026-03-16
iOS Advertised Solicited Services


ajr33:
**Describe the bug**
On iOS, Solicited Services are not shown in the advertising packet but in a comment, they look like they were tested at one point but not implemented (shown in ParseAdvertismentData on line 296). Android shows up fine with the solicited services.

**To Reproduce**
1. Advertise a device with a solicited service.
2. Use the Plugin.BLE library to scan for the associated device on iOS.
3. Examine the advertisement records and you will see no solicited service record listed. 

**Expected behavior**
An advertisement record of the solicited service is listed with the discovery of a device.

**Configuration (please complete the following information):**
 - Device: iPhone 11
 - OS: iOS 18.2.1
 - Plugin.BLE Version: 3.1.0

**Additional context**
I would like to use this advertisement data to know the state of a BLE device I am working on. I could put this data in a characteristic or use a normal advertisement for a service uuid but this should be possible on iOS. CBAdvertisementDataSolicitedServiceUUIDsKey seems to be the missing attribute to parse.

janusw:
So the suggestion here is to extend `ParseAdvertismentData` to handle `CBAdvertisement.DataSolicitedServiceUUIDsKey`. Should certainly be possible and could be done analogous to `DataServiceUUIDsKey` and `DataOverflowServiceUUIDsKey`. PR welcome 😉

janusw:
> So the suggestion here is to extend `ParseAdvertismentData` to handle `CBAdvertisement.DataSolicitedServiceUUIDsKey`. Should certainly be possible and could be done analogous to `DataServiceUUIDsKey` and `DataOverflowServiceUUIDsKey`. PR welcome 😉

@copilot Please open a PR to implement this ...

ajr33:
The copilot suggestion works quite well as this is a very simple change to include and it works on the device I was previously testing on. Do you need me to do anymore for this?

janusw:
> The copilot suggestion works quite well as this is a very simple change to include and it works on the device I was previously testing on.

Awesome, thanks for testing!


> Do you need me to do anymore for this?

No, if you tested it and it works well for you, then we can merge it (and it will be part of the next release, of course).

ajr33:
Okay that sounds good. Thanks for quick response for this!

janusw:
[...]

PR #989, opened 2026-03-21
Fix iOS 32 bit UUIDs missing from advertisement records.

ajr33:
This pull request is in continuation of testing from #985. 
Fix iOS 32 bit UUIDs missing from advertisement record due to a typo with the length case. It was set to 8 bytes instead of 4 for 32 bit UUIDs.
Added backing property for sample to see all advertisement records when change/updated.

### DIFF
--- a/Source/Plugin.BLE/Apple/Adapter.cs
+++ b/Source/Plugin.BLE/Apple/Adapter.cs
@@ -336,7 +336,7 @@ namespace Plugin.BLE.iOS
                                 // 128-bit UUID
                                 records.Add(new AdvertisementRecord(AdvertisementRecordType.UuidsComplete128Bit, cbuuid.Data.ToArray()));
                                 break;
-                            case 8:
+                            case 4:
                                 // 32-bit UUID
                                 records.Add(new AdvertisementRecord(AdvertisementRecordType.UuidCom32Bit, cbuuid.Data.ToArray()));
                                 break;
@@ -364,7 +364,7 @@ namespace Plugin.BLE.iOS
                                 // 128-bit solicited service UUID
                                 records.Add(new AdvertisementRecord(AdvertisementRecordType.SsUuids128Bit, cbuuid.Data.ToArray()));
                                 break;
-                            case 8:
+                            case 4:
                                 // 32-bit solicited service UUID
                                 records.Add(new AdvertisementRecord(AdvertisementRecordType.SsUuids32Bit, cbuuid.Data.ToArray()));
                                 break;
--- a/Source/BLE.Client/BLE.Client.Maui/ViewModels/BLEDeviceViewModel.cs
+++ b/Source/BLE.Client/BLE.Client.Maui/ViewModels/BLEDeviceViewModel.cs
@@ -76,7 +76,20 @@ namespace BLE.Client.Maui.ViewModels
             }
         }
 
-        public IReadOnlyList<AdvertisementRecord> AdvertisementRecords;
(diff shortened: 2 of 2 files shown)


janusw:
> Fix iOS 32 bit UUIDs missing from advertisement record due to a typo with the length case. It was set to 8 bytes instead of 4 for 32 bit UUIDs.

Good catch! Thanks for the contribution!

It seems that in PR #989 this problem was inherited from the existing code, which was contributed in #446. I shortly wondered whether there could be a particular reason for this irregularity, but since nothing is mentioned in this regard in #445 or #446, I assume it was simply a typo.

Source/Plugin.BLE/Apple/Adapter.cs
```diff
@@ -336,7 +336,7 @@ public static List<AdvertisementRecord> ParseAdvertismentData(NSDictionary adver
                                 // 128-bit UUID
                                 records.Add(new AdvertisementRecord(AdvertisementRecordType.UuidsComplete128Bit, cbuuid.Data.ToArray()));
                                 break;
-                            case 8:
+                            case 4:
                                 // 32-bit UUID
                                 records.Add(new AdvertisementRecord(AdvertisementRecordType.UuidCom32Bit, cbuuid.Data.ToArray()));
```
janusw, line 341: I just noticed another small irregularity: The naming of values in the `AdvertisementRecordType` enum is inconsistent in some cases.

My suggestion would be to make the following modifications:
* `UuidCom32Bit` => `UuidComplete32Bit`
* `UuidsIncomple16Bit` => `UuidsIncomplete16Bit`

@ajr33 Could you include this small change in this PR?

ajr33: Yeah I can do that. I also saw other inconsistencies with the following:

- (0x16) ServiceData -> ServiceDataUuid16Bit
- (0x21) ServiceData128Bit -> ServiceDataUuid128Bit

Would adding these changes be okay or not? I am thinking about how code may break with this but if a user pushes an update they should be able to see the modifications. 

Thoughts?

janusw: Well, ok, if there are even more of those, I'll merge this one now and will take care of the renaming in a separate PR. And you're right, this will potentially break the API, so we have to be a bit careful about it ...

janusw (approved):

Source/Plugin.BLE/Apple/Adapter.cs
ajr33: Okay sounds good. 
Is there a way for me to have these changes in a pre-release? I would like to be able to use it from NuGet instead of pulling the whole project into my application project. If that is unavoidable then I can figure it out with visual studio. I'm not sure if there is a nightly build or something that can be done.

janusw:
> Okay sounds good. Is there a way for me to have these changes in a pre-release? I would like to be able to use it from NuGet instead of pulling the whole project into my application project.

@ajr33 Yes, there is. Our CI (Github Actions) builds a nuget package for basically every commit. The one for the merge commit of this PR can be found here: https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/actions/runs/23451527921#artifacts

The version of the package is: `3.3.0-beta.1-5-g94b75ba`

You can download this and use it locally in your project, e.g. via a `nuget.config` file with `<packageSources>` (see https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#packagesources).

Issue #986, opened 2026-03-19
When Android is Limited or Disconnected DisconnectDeviceAsync hangs.


ajr33:
**Describe the bug**
When trying to disconnect from a BLE device on an android phone with the state of the device as Limited or Disconnected, the method _adapter.DisconnectedDeviceAsync hangs forever.

**To Reproduce**
1. Connect to a device via _adapter.ConnectToDeviceAsync
2. Make some simple GATT calls (no auth required)
3. Attempt to disconnect the device.

**Expected behavior**
It would be expected to either have the method return immediately or throw an exception if trying to disconnect a device that is already disconnected. Freezing up in the await call is definitely not desirable.

**Configuration (please complete the following information):**
 - Device: Pixel 6a
 - OS: Android 16
 - Plugin.BLE Version:  3.1.0

**Additional context**
I saw this problem when trying to connect to a device, retrieve some basic information (Device Information Service properties) and then immediately disconnect the device afterwards. Not sure how the underlying device is getting set to Limited connection, which is causing the problem mainly, but I also seen it happen with the state set to Disconnected before calling DisconnectedDeviceAsync. iOS does not have this issue. A simple workaround I can do now is check if the state is Limited or Disconnected and skip disconnection if it is one of these two, the only problem I have with that is that I am not sure whether or not all the states and objects that need to be discarded will be since the device I am connecting to says that it is still connected to the phone. I have another workaround that can work in my specific instance but would not for others.

ajr33:
After further research, I realized that I was connecting to a new advertised device (not necessary a new device) which causes the limited to show since between connecting there are more advertisement packets that happen in between. Therefore connecting and GATT operations work but since I was trying to disconnect from the limited device that is the problem. I did fix this issue in my code but this issue still remains since disconnecting from a limited device should not hang the API call which is a bit confusing to me still as to why this happens. This weekend I may be able to take a look.

ajr33:
Did some more testing today and found that this code...

```c#
return TaskBuilder.FromEvent<bool, EventHandler<DeviceEventArgs>, EventHandler<DeviceErrorEventArgs>>(
   execute: () => DisconnectDeviceNative(device),

   getCompleteHandler: (complete, reject) => ((sender, args) =>
   {
       if (args.Device.Id == device.Id)
       {
           Trace.Message("DisconnectAsync Disconnected: {0} {1}", args.Device.Id, args.Device.Name);
           complete(true);
       }
[...]

PR #993, opened 2026-04-21
Fix Android hanging on disconnection when status is Limited

ajr33:
This pull request addresses issue: #986.

When the gatt instance is null, calling DisconnectDeviceAsync will hang forever. This seems to happen if a device with status "Limited" is passed in as the parameter since it is not fully connected (hence why gatt instance is null) the call should fail but currently does not. This pull request simply adds an exception that is thrown when the gatt instance is null. This fixes the thread hanging forever and also allows for the uses to debug and determine why the call failed.

### DIFF
--- a/Source/Plugin.BLE/Android/Device.cs
+++ b/Source/Plugin.BLE/Android/Device.cs
@@ -14,6 +14,7 @@ using System.Threading;
 using Java.Util;
 using Plugin.BLE.Extensions;
 using Plugin.BLE.Abstractions.Extensions;
+using Plugin.BLE.Abstractions.Exceptions;
 
 namespace Plugin.BLE.Android
 {
@@ -212,7 +213,8 @@ namespace Plugin.BLE.Android
             else
             {
                 Trace.Message("[Warning]: Can't disconnect {0}. Gatt is null.", Name);
-            }
+				throw new DeviceConnectionException(Id, Name, $"Cannot disconnect device '{Name}': device is not properly connected (no GATT instance). Current state: {State}.");
+			}
         }
 
         /// <summary>
--- a/Source/Plugin.BLE/Android/Device.cs
+++ b/Source/Plugin.BLE/Android/Device.cs
@@ -107,11 +107,11 @@ namespace Plugin.BLE.Android
 
         private async Task<IReadOnlyList<IService>> DiscoverServicesInternal(CancellationToken cancellationToken)
         {
-	        if (_gatt == null)
-	        {
-		        Trace.Message("[Warning]: Can't discover services {0}. Gatt is null.", Name);
-	        }
-	        
+            if (_gatt == null)
+            {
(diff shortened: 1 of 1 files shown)


janusw (commented):
> This pull request simply adds an exception that is thrown when the gatt instance is null. This fixes the thread hanging forever and also allows for the uses to debug and determine why the call failed.

It seems to be quite similar to Copilot's PR #992, but there are two small differences AFAICS:
* You are using a `DeviceConnectionException` instead of an `InvalidOperationException`. This makes sense, since the problem affects the connection (or better: disconnection) of a device.
* In contrast to Copilot, you are keeping the trace message, instead of replacing it with the exception. Any particular reason for that? I'd also tend to remove it, since the exception should carry all relevant information.

Source/Plugin.BLE/Android/Device.cs
```diff
@@ -212,7 +213,8 @@ public void Disconnect()
             else
             {
                 Trace.Message("[Warning]: Can't disconnect {0}. Gatt is null.", Name);
-            }
+				throw new DeviceConnectionException(Id, Name, $"Cannot disconnect device '{Name}': device is not properly connected (no GATT instance). Current state: {State}.");
+			}
```
janusw: One nitpick: Your indentation here is off, or rather depends on the editor settings, since you sem to be indenting with tabs. Please use spaces instead (as in the surrounding lines).

janusw:
Note that our CI build fails on MacOS with the error:
```
This version of .NET for MacCatalyst (26.2.10233) requires Xcode 26.3.
The current version of Xcode is 26.2.
Either install Xcode 26.3, or use a different version of .NET for MacCatalyst.
See https://aka.ms/xcode-requirement for more information.
```
This is not caused by your changes, of course, but rather by the CI setup (probably wel'll just need to bump the XCode version). I'll take care of this ...

Source/Plugin.BLE/Android/Device.cs
ajr33: Could you advise how'd you caught this? I am looking at SourceTree and still don't see a difference. Visual Studio said mixed and shows these lines (as well as some others) that were using tabs but I don't see the indentation differences at all.
Updates coming in soon.

janusw: Actually the indentation also looks well for me when viewing the code with VSCode, for example, which seems to be set up to have a tab width of four chars. However, I often use gitk for visualizing the git commit tree and looking at diffs, and this shows a wrong indentation of your lines (due to a eight-char tab width).

Mixing space indentation with tab indentation is a bad idea, because the display of tab-indented lines depends on the tab width set up in the editor/IDE.

janusw (approved):
The CI failure is gone after rebasing and the changes look good to me by now. Thanks for the contribution!

PR #994, opened 2026-04-22
fix #877 - add option to ConnectParameters CheckIsLeDeviceType? and o…

### LINKED ISSUE #877: Bug: [Adapter] Device {..} does not support BLE
We have recently updated to `3.1.0`, but unfortunately this version broke our code, because of a new check that has been introduced.

Our usage looks something like that:
```c#
var adapter = CrossBluetoothLE.Current.Adapter;
var device = await adapter.ConnectToKnownDeviceAsync(new Guid(guid), new ConnectParameters(forceBleTransport: true), tokenSource.Token);
``´
More specifically the problem is that on newer android devices the `BluetoothDevice.Type` is `BluetoothDeviceType.Unkown`, and this is also true if the ble device isn't even in Range. This causes the exception you can find in this snipped
[https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/2c7972e4ec3f41654b6422d7cec6b8ca72694d38/Source/Plugin.BLE/Android/Adapter.cs#L310](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/2c7972e4ec3f41654b6422d7cec6b8ca72694d38/Source/Plugin.BLE/Android/Adapter.cs#L310)

Here the implementation of SupportsBLE which uses the `Type` property:
[https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/2c7972e4ec3f41654b6422d7cec6b8ca72694d38/Source/Plugin.BLE/Android/Extensions/BluetoothDeviceExtension.cs#L7](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/2c7972e4ec3f41654b6422d7cec6b8ca72694d38/Source/Plugin.BLE/Android/Extensions/BluetoothDeviceExtension.cs#L7)


We need this check to be removed, instead we cannot upgrade to the next version.

### PR DESCRIPTION
**Pull request by alex3696**
…ptional check NativeDevice.SupportsBLE before ConnectGatt

### DIFF
--- a/Source/Plugin.BLE/Android/Adapter.cs
+++ b/Source/Plugin.BLE/Android/Adapter.cs
@@ -258,8 +258,6 @@ namespace Plugin.BLE.Android
             var nativeDevice = _bluetoothAdapter.GetRemoteDevice(macBytes);
             if (nativeDevice == null)
                 throw new Abstractions.Exceptions.DeviceConnectionException(deviceGuid,"", $"[Adapter] Device {deviceGuid} not found.");
-            if (!nativeDevice.SupportsBLE())
-                throw new Abstractions.Exceptions.DeviceConnectionException(deviceGuid,"", $"[Adapter] Device {deviceGuid} does not support BLE.");
             var device = new Device(this, nativeDevice, null);
 
             await ConnectToDeviceAsync(device, connectParameters, cancellationToken);
--- a/Source/Plugin.BLE/Android/Device.cs
+++ b/Source/Plugin.BLE/Android/Device.cs
@@ -8,6 +8,7 @@ using Android.Content;
 using Plugin.BLE.Abstractions;
 using Plugin.BLE.Abstractions.Contracts;
 using Plugin.BLE.Abstractions.Utils;
+using Plugin.BLE.Android.Extensions;
 using Plugin.BLE.Android.CallbackEventArgs;
 using Trace = Plugin.BLE.Abstractions.Trace;
 using System.Threading;
@@ -146,6 +147,9 @@ namespace Plugin.BLE.Android
         {
             IsOperationRequested = true;
             ConnectParameters = connectParameters;
+
+            if (connectParameters.CheckIsLeDeviceType && !NativeDevice.SupportsBLE())
+                throw new Abstractions.Exceptions.DeviceConnectionException(Id, "", $"[Adapter] Device {Id} does not support BLE.");
 
             if (connectParameters.ForceBleTransport)
             {
--- a/Source/Plugin.BLE/Android/Device.cs
+++ b/Source/Plugin.BLE/Android/Device.cs
@@ -150,7 +150,7 @@ namespace Plugin.BLE.Android
             ConnectParameters = connectParameters;
 
             if (connectParameters.CheckIsLeDeviceType && !NativeDevice.SupportsBLE())
-                throw new Abstractions.Exceptions.DeviceConnectionException(Id, "", $"[Adapter] Device {Id} does not support BLE.");
+                throw new DeviceConnectionException(Id, Name, $"[Device] Device {Id} does not support BLE, type is {NativeDevice.Type}. You can disable this check, see more info in ConnectParameters.CheckIsLeDeviceType description.");
 
             if (connectParameters.ForceBleTransport)
             {
--- a/Source/Plugin.BLE/Shared/ConnectParameter.cs
+++ b/Source/Plugin.BLE/Shared/ConnectParameter.cs
@@ -16,6 +16,12 @@
         /// </summary>
         public bool ForceBleTransport { get; }
 
+		/// <summary>
+        /// Android only: Сheck and connect only if the BluetoothDeviceType is Le or Dual. 
+		/// If the BluetoothDeviceType is Unknown or Classic, a DeviceConnectionException will be thrown.
+        /// </summary>
+        public bool CheckIsLeDeviceType { get; } = true;
+
         /// <summary>
         /// Windows only, mapped to:
         /// https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothlepreferredconnectionparameters
--- a/Source/Plugin.BLE/Shared/ConnectParameter.cs
+++ b/Source/Plugin.BLE/Shared/ConnectParameter.cs
@@ -17,11 +17,13 @@
         public bool ForceBleTransport { get; }
 
         /// <summary>
-        /// Android only: Сheck and connect only if the BluetoothDeviceType is Le or Dual. 
-        /// If the BluetoothDeviceType is Unknown or Classic, a DeviceConnectionException will be thrown.
-        /// If the BluetoothDeviceType is Unknown - try scanning (or wait for the cache to load after rebooting the device) 
+        /// Android only: Strict BluetoothDeviceType checking.
+        /// The connection will only be attempted if the device supports LE, otherwise a DeviceConnectionException will be thrown.
+        /// This check is an early “warning” of what might happen next - error GATT 133 or Bluetooth stack fault.
+        /// The BluetoothDeviceType may be Unknown immediately after the device is rebooted, or if the type is not accepted correctly, try scanning to get or update the type.
+        /// If the device intentionally does not declare the connection type, you can disable this check.
         /// </summary>
-        public bool CheckIsLeDeviceType { get; } = true;
+        public bool CheckIsLeDeviceType { get; set; } = true;
 
         /// <summary>
         /// Windows only, mapped to:
--- a/Source/Plugin.BLE/Shared/ConnectParameter.cs
+++ b/Source/Plugin.BLE/Shared/ConnectParameter.cs
@@ -15,7 +15,7 @@
         /// Android only: For Dual Mode device, force transport mode to LE. The default is false.
         /// </summary>
         public bool ForceBleTransport { get; }
-        
+
         /// <summary>
         /// Android only: Сheck and connect only if the BluetoothDeviceType is Le or Dual. 
         /// If the BluetoothDeviceType is Unknown or Classic, a DeviceConnectionException will be thrown.
--- a/Source/Plugin.BLE/Shared/ConnectParameter.cs
+++ b/Source/Plugin.BLE/Shared/ConnectParameter.cs
@@ -15,10 +15,11 @@
         /// Android only: For Dual Mode device, force transport mode to LE. The default is false.
         /// </summary>
         public bool ForceBleTransport { get; }
-
-		/// <summary>
+        
+        /// <summary>
         /// Android only: Сheck and connect only if the BluetoothDeviceType is Le or Dual. 
-		/// If the BluetoothDeviceType is Unknown or Classic, a DeviceConnectionException will be thrown.
+        /// If the BluetoothDeviceType is Unknown or Classic, a DeviceConnectionException will be thrown.
+        /// If the BluetoothDeviceType is Unknown - try scanning (or wait for the cache to load after rebooting the device) 
         /// </summary>
         public bool CheckIsLeDeviceType { get; } = true;
 
--- a/Source/Plugin.BLE/Shared/ConnectParameter.cs
+++ b/Source/Plugin.BLE/Shared/ConnectParameter.cs
@@ -3,18 +3,18 @@
     /// <summary>
     /// Connection parameters. Contains platform specific parameters needed to achieved connection
     /// </summary>
-    public struct ConnectParameters
+    public readonly struct ConnectParameters
     {
         /// <summary>
         /// Android only, from documentation:  
         /// boolean: Whether to directly connect to the remote device (false) or to automatically connect as soon as the remote device becomes available (true).
         /// </summary>
-        public bool AutoConnect { get; }
+        public bool AutoConnect { get; init; } = false;
 
         /// <summary>
         /// Android only: For Dual Mode device, force transport mode to LE. The default is false.
         /// </summary>
-        public bool ForceBleTransport { get; }
+        public bool ForceBleTransport { get; init; } = false;
 
         /// <summary>
         /// Android only: Strict BluetoothDeviceType checking.
@@ -23,13 +23,13 @@
         /// The BluetoothDeviceType may be Unknown immediately after the device is rebooted, or if the type is not accepted correctly, try scanning to get or update the type.
         /// If the device intentionally does not declare the connection type, you can disable this check.
         /// </summary>
-        public bool CheckIsLeDeviceType { get; set; } = true;
+        public bool CheckIsLeDeviceType { get; init; } = false;
 
         /// <summary>
         /// Windows only, mapped to:
         /// https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothlepreferredconnectionparameters
         /// </summary>
-        public ConnectionParameterSet ConnectionParameterSet { get; }
+        public ConnectionParameterSet ConnectionParameterSet { get; init; } = ConnectionParameterSet.None;
 
         /// <summary>
         /// Default-constructed connection parameters (all parameters set to false).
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by janusw** [comment]
@KevinJosephMorin Can you test this PR and tell us if it works for your purpose? There is a nupkg generated in the GitHub Actions builld that you can use ...

**Comment by janusw** [comment]
> @KevinJosephMorin Can you test this PR and tell us if it works for your purpose? There is a nupkg generated in the GitHub Actions builld that you can use ...

Or anyone else who was bitten by #877 ...?  @BierDav @IchoTolot @AskBojesen

File: Source/Plugin.BLE/Shared/ConnectParameter.cs [hunk 1]
```diff
@@ -16,6 +16,12 @@ public struct ConnectParameters
         /// </summary>
         public bool ForceBleTransport { get; }
 
+		/// <summary>
+        /// Android only: Сheck and connect only if the BluetoothDeviceType is Le or Dual. 
+		/// If the BluetoothDeviceType is Unknown or Classic, a DeviceConnectionException will be thrown.
+        /// </summary>
+        public bool CheckIsLeDeviceType { get; } = true;
```
janusw: One minor nitpick: You're partly indenting with tabs here and partly with spaces. Please use only spaces. 🙏  Tabs can lead to messed-up indentation (depending on editor/IDE settings) ...

Source/Plugin.BLE/Android/Device.cs
```diff
@@ -148,6 +149,9 @@ public void Connect(ConnectParameters connectParameters, CancellationToken cance
             IsOperationRequested = true;
             ConnectParameters = connectParameters;
 
+            if (connectParameters.CheckIsLeDeviceType && !NativeDevice.SupportsBLE())
+                throw new Abstractions.Exceptions.DeviceConnectionException(Id, "", $"[Adapter] Device {Id} does not support BLE.");
```
janusw: What is actually the reason that you move the exception here from `Adapter.ConnectToKnownDeviceNativeAsync`?

alex3696: 1) In case of connection, the NativeDevice.SupportsBLE check applies specifically to the Device, it is more logical to place it there.
2) This will ensure uniform behavior of all connection functions and additionally protect when receiving a partial advertising package, in this scenario:
```
btDevice = await adapter.DiscoverDeviceAsync(guid, ct);
if (null != btDevice)
    await adapter.ConnectToDeviceAsync(device, conn_param, cts.Token);
```
My observations show that manufacturers of different devices implement drivers and Bluetooth communication procedures differently.
For example, when BluetoothDeviceType.Unkown, connectGatt crashes on some with error GATT 133 and that’s it. At the same time, on others it is determined that the device supports LE and the connection is successful,
perhaps using “internal scanning” or the fact of a successful connection. Recently, the behavior of Bluetooth devices has become more stable, and I receive fewer comments. In my code I use a single connection function with the following logic:
```
_btDevice = await ConnectToPairedAsync(_Adapter, guid, conn_param, ct);
if (null == _btDevice)
{
    _btDevice = await ConnectToKnownAsync(_Adapter, guid, conn_param, ct);
    if (null == _btDevice)
    {
    	_btDevice = await ConnectToScannedAsync(_Adapter, guid, conn_param, ct);
    }
}
if (null == _btDevice)
    throw new Exception("Can`t get device ");
```

Source/Plugin.BLE/Shared/ConnectParameter.cs
alex3696: I didn't think about it :-) : partly copied from you, partly added in my VS IDE, which has Tabs by default :-)

Source/Plugin.BLE/Android/Device.cs
janusw: > 1. In case of connection, the NativeDevice.SupportsBLE check applies specifically to the Device, it is more logical to place it there.

This makes sense (in principle) ...

>  2. This will ensure uniform behavior of all connection functions

... but this point is the more crucial one, I think. In particular the check is now also applied to `ConnectToDeviceAsync` (not only to `ConnectToKnownDeviceAsync`).

As you argue, this assures more uniform behavior, but it also changes the behavior of `ConnectToDeviceAsync`. And since the check already made some trouble in `ConnectToKnownDeviceAsync` for several people (see the reports in #877), I'm afraid it might make even more trouble when applied in `ConnectToDeviceAsync`.

I'd rather fix the regression of #877 than introduce new ones. So, if we really want to go that route, I'd propose to:
* either change the default value of `CheckIsLeDeviceType` to `false` (in order to restore the default behavior of version 3.0 and earlier, and let people who need stricter checking enable it manually).
* or at least mention the `CheckIsLeDeviceType` parameter in the exception that is thrown by the check, in order to give people (who have trouble with this) a pointer how to overcome the problem.

> My observations show that manufacturers of different devices implement drivers and Bluetooth communication procedures differently.

Yes, this is unfortunately a situation we have to deal with. We should try to set up the library such that the default behavior works for as many people as possbile.

As written above, I'd really like to have some feedback from the sufferers of #877, before I approve this PR. @KevinJosephMorin @BierDav @IchoTolot @AskBojesen

janusw:
> > @KevinJosephMorin Can you test this PR and tell us if it works for your purpose? There is a nupkg generated in the GitHub Actions builld that you can use ...
> 
> Or anyone else who was bitten by #877 ...? @BierDav @IchoTolot @AskBojesen

The nupkg is at https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/actions/runs/24872433574#artifacts. I would be grateful if someone could test it (and/or contribute to the discussion in this PR).

Source/Plugin.BLE/Android/Device.cs
alex3696: > I'd rather fix the regression of #877 than introduce new ones.

I understand your concerns, but if it were a regression, it would be a massive problem - while the situation is unclear (3 people from (3.1)452 811+(3.2)53 652 downloads) . I think we'll have to wait for detailed feedback - with what fun peripherals, on what versions of Android, on what phones and under what conditions.  
@AskBojesen wrote, "I have not encountered this problem for at least 10 phones with Android 14/15 = API 34/35" - https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/issues/877#issuecomment-3424203940. 

I described the problem in detail and proposed a solutions. Checking is just an early “warning” of what might happen next - error GATT 133 or stack fault or break the device. Windows and iOS do not allow such connections by design. I will be happy with any option (CheckIsLeDeviceType = false | true). 
What are your preferences?

IchoTolot:
It’s been a while since I worked on this issue. 
As I recall, starting with a certain API version, Android would always return `false` for nativeDevice.SupportsBLE() when attempting to connect without a prior call to StartScanningForDevicesAsync().

I ignored that as a test, but that resulted in error code 133.

In the end, I solved it for myself by always running a scan before a connection attempt. 
While this delays the connection setup, it works reliably.

KevinJosephMorin:
> > > @KevinJosephMorin Can you test this PR and tell us if it works for your purpose? There is a nupkg generated in the GitHub Actions builld that you can use ...
> > 
> > 
> > Or anyone else who was bitten by #877 ...? @BierDav @IchoTolot @AskBojesen
> 
> The nupkg is at https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/actions/runs/24872433574#artifacts. I would be grateful if someone could test it (and/or contribute to the discussion in this PR).

@janusw , This thread was in my spam.  Sorry I didn't see it earlier.  I will be near a device later today and will test.

KevinJosephMorin:
I just tried to update the Nuget and got an error.  Is there something extra that needs to be done to support MonoAndroid,Version=v13.0

PM> Install-Package Plugin.BLE -Version 3.3.0-beta.1-49-g56d70e2 -Prerelease
Install-Package : NU1202: Package Plugin.BLE 3.3.0-beta.1-49-g56d70e2 is not compatible with monoandroid13.0 (MonoAndroid,Version=v13.0). Package Plugin.BLE 3.3.0-beta.1-49-g56d70e2 supports:
  - net10.0 (.NETCoreApp,Version=v10.0)
  - net10.0-android36.0 (.NETCoreApp,Version=v10.0)
  - net10.0-ios26.0 (.NETCoreApp,Version=v10.0)
  - net10.0-maccatalyst26.0 (.NETCoreApp,Version=v10.0)
  - net10.0-windows10.0.19041 (.NETCoreApp,Version=v10.0)
  - net9.0 (.NETCoreApp,Version=v9.0)
  - net9.0-android35.0 (.NETCoreApp,Version=v9.0)
  - net9.0-ios18.0 (.NETCoreApp,Version=v9.0)
  - net9.0-maccatalyst18.0 (.NETCoreApp,Version=v9.0)
  - net9.0-windows10.0.19041 (.NETCoreApp,Version=v9.0)
At line:1 char:1
+ Install-Package Plugin.BLE -Version 3.3.0-beta.1-49-g56d70e2 -Prerele ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Install-Package], Exception
    + FullyQualifiedErrorId : NuGetCmdletUnhandledException,NuGet.PackageManagement.PowerShellCmdlets.InstallPackageCommand

janusw:
> I just tried to update the Nuget and got an error. Is there something extra that needs to be done to support MonoAndroid,Version=v13.0

Are you still using Xamarin? In the 3.3 beta, we have actually removed Xamarin support and only support recent .NET versions with MAUI ...

KevinJosephMorin:
I am still using Xamarin for a few more months.  Should I stick with 3.0.0?

janusw:
> I am still using Xamarin for a few more months. Should I stick with 3.0.0?

Well, since #877 is basically a regression in v3.1, I'd consider backporting the fix to v3.1 and/or v3.2 (both of which still support Xamarin).

Unfortunately this PR cannot be tested with Xamarin, but I think I can create a branch with these fixes based on v3.2, so that you can test it. (Will do that tonight ...)

janusw:
> > I am still using Xamarin for a few more months. Should I stick with 3.0.0?
> 
> Unfortunately this PR cannot be tested with Xamarin, but I think I can create a branch with these fixes based on v3.2, so that you can test it.

I have backported the commits in this PR to this branch now (for testing with Xamarin): https://github.com/janusw/dotnet-bluetooth-le/commits/backport_pr994_to_3.2/

However, I think some of Copilot's comments may need to be fixed first, before it is fully usable ...

alex3696:
I was in a hurry. It's strange that no one noticed that the parameter is immutable - fixed. These changes need to be backported to 3.2
Changed the description and exception message to quickly understand the problem.

janusw:
> I was in a hurry. It's strange that no one noticed that the parameter is immutable - fixed.

Thanks for fixing 👍. I guess it was not noticed because no one seriously tried to use it yet 😉 (and I simply did not look close enough) ...


> These changes need to be backported to 3.2

Done ✅

KevinJosephMorin:
@janusw Is there a build you would like me to test on Xamarin?

janusw:
> @janusw Is there a build you would like me to test on Xamarin?

@KevinJosephMorin  The nupkg for the backported fix is here: https://github.com/janusw/dotnet-bluetooth-le/actions/runs/25003142375#artifacts
Would be great if you could give it a try ...

KevinJosephMorin:
> > @janusw Is there a build you would like me to test on Xamarin?
> 
> @KevinJosephMorin The nupkg for the backported fix is here: https://github.com/janusw/dotnet-bluetooth-le/actions/runs/25003142375#artifacts Would be great if you could give it a try ...

That works for me.  I upgraded and confirmed it didn't work.  Added the new objParameters.CheckIsLeDeviceType = false, and then it works.  Are you think you will release 3.2.1?

janusw:
> > @KevinJosephMorin The nupkg for the backported fix is here: https://github.com/janusw/dotnet-bluetooth-le/actions/runs/25003142375#artifacts Would be great if you could give it a try ...
> 
> That works for me. I upgraded and confirmed it didn't work. Added the new objParameters.CheckIsLeDeviceType = false, and then it works.

Awesome, thanks for checking!


> Are you think you will release 3.2.1?

I will!

janusw:
> > Are you think you will release 3.2.1?
> 
> I will!

... but first let's get this into PR into main! After that I will backport it to 3.2 and preprepare a new release there. Also I plan to release a second beta for 3.3.0 soon.

Source/Plugin.BLE/Android/Device.cs
janusw: > I will be happy with any option (CheckIsLeDeviceType = false | true). What are your preferences?

I would have a slight tendency for using a default of `false`, in order to stick to the pre-3.1 behavior by default, and allow to do stricter checking if needed. Simply because the check makes trouble for some people.

OTOH, since we now refer to the checking option in the exception text, it might be easier to get rid of the exception when doing the check than dealing with the strange errors one can run into without the check.

I have no feeling of how widespread the problem with GATT errror 133 or stack faults is. AFAICS you were the only one reporting it for some rather exotic China phones ...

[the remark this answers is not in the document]
janusw: This is fixed.

Source/Plugin.BLE/Shared/ConnectParameter.cs
```diff
@@ -16,6 +16,13 @@ public struct ConnectParameters
         /// </summary>
         public bool ForceBleTransport { get; }
 
+        /// <summary>
+        /// Android only: Сheck and connect only if the BluetoothDeviceType is Le or Dual.
```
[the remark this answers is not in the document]
janusw: Fixed, I think.

Source/Plugin.BLE/Shared/ConnectParameter.cs
```diff
@@ -16,6 +16,13 @@ public struct ConnectParameters
         /// </summary>
         public bool ForceBleTransport { get; }
 
+        /// <summary>
+        /// Android only: Сheck and connect only if the BluetoothDeviceType is Le or Dual. 
+        /// If the BluetoothDeviceType is Unknown or Classic, a DeviceConnectionException will be thrown.
+        /// If the BluetoothDeviceType is Unknown - try scanning (or wait for the cache to load after rebooting the device) 
+        /// </summary>
+        public bool CheckIsLeDeviceType { get; } = true;
```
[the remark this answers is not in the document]
janusw: This has been addressed by making `CheckIsLeDeviceType` settable, but that is actually inconsistent with the other members of the `ConnectParameters` struct, which are read-only.

I'd propose to primarily:
* Add an constructor parameter for `CheckIsLeDeviceType` (for consistency with the other options).

And optionally:
* Use an `init` setter for `CheckIsLeDeviceType` and all other members of `ConnectParameters`.

That would be more consistent than only adding a setter to `CheckIsLeDeviceType`.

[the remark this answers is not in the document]
janusw: Implemented both on the branch in a13248e414acf6112c371d4d2647785323a1cdd3.

Source/Plugin.BLE/Android/Adapter.cs
```diff
@@ -258,8 +258,6 @@ public override async Task<IDevice> ConnectToKnownDeviceNativeAsync(Guid deviceG
             var nativeDevice = _bluetoothAdapter.GetRemoteDevice(macBytes);
             if (nativeDevice == null)
                 throw new Abstractions.Exceptions.DeviceConnectionException(deviceGuid,"", $"[Adapter] Device {deviceGuid} not found.");
```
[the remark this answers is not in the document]
janusw: This should also be fixed via a13248e414acf6112c371d4d2647785323a1cdd3. `CheckIsLeDeviceType` is false by default (in the struct and int the constructor). One has to enable it explicitly to activate the check.

Source/Plugin.BLE/Shared/ConnectParameter.cs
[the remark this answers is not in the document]
janusw: Actually I think one can fully remove the constructor, now that all members are init-settable. I don't want to do it in this PR, though, because I want to backport it to 3.2 (and therefore avoid incompatible changes).

I will remove the constructor in a follow-up (and therefore refrain from adding a unit test for it) ...

janusw:
From my point of view, this PR is in a mergable state now. I will wait a bit, in order to allow people to comment, and merge it to main within a week, if no one objects ...
@alex3696 @KevinJosephMorin @IchoTolot

janusw (approved):

PR #1004, opened 2026-08-21
Support api level 33 overrides for memory save characteristic values

mathiasnielsen:
On Android API level 33+, there are two overrides that provide the updated characteristic value. Using these values ensures the data remains memory-safe, as reading directly from the characteristic itself may result in the value changing too quickly.

### DIFF
--- a/Source/Plugin.BLE/Android/CallbackEventArgs/CharacteristicReadCallbackEventArgs.cs
+++ b/Source/Plugin.BLE/Android/CallbackEventArgs/CharacteristicReadCallbackEventArgs.cs
@@ -7,10 +7,13 @@ namespace Plugin.BLE.Android.CallbackEventArgs
         public BluetoothGattCharacteristic Characteristic { get; }
         public GattStatus Status { get; }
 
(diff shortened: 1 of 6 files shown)


janusw:
Thanks for the contribution. Note that the pipeline failure on MacOS is a general CI problem (not your fault!). I'm going to take care of this ...

CodeMusher:
What about the Windows build?

janusw:
> Note that the pipeline failure on MacOS is a general CI problem (not your fault!). I'm going to take care of this ...

@mathiasnielsen I have fixed this problem on the main branch. Could you please merge that fix into your branch (or alternatively rebase your branch on current main)? I tried to do it myself, but apparently I lack the permissions to update your branch.

janusw:
> What about the Windows build?

The Windows build seems to fail because the `CharacteristicUpdatedEventArgs` class in the shared code is changed, but the Windows-specific code has not been updated accordingly (in contrast to the other platforms).

@mathiasnielsen This also needs to be fixed.

